PR #20, opened 2022-01-24
Add Github workflow for running tests

joycse06:
Add Github workflow for running tests


daniel-nichter (approved):

daniel-nichter:
Going to merge despite failures because they're due to not having MySQL. So I'll fix tests in another PR to `SkipAll` if MySQL isn't running (until we get MySQL running in Actions).

PR #25, opened 2022-02-10
Bugfix to log and err handler

jemiahw:
reload_password.go was writing the password to debug logs. Now redacted.

Fixed a bug in engine. Prepare err handler, where it referenced the wrong err, masking the actual error.


daniel-nichter (approved):

daniel-nichter:
Excellent, thank you!

PR #34, opened 2022-05-11
plan/sort: filter out dupe metrics during level metric aggregation

allanliu:
Fixes bug in `plan.Sort` where duplicate domain metrics from a higher frequency level are also aggregated into the lower frequency level.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,6 @@ require (
 	github.com/signalfx/golib/v3 v3.3.36
 	github.com/stretchr/testify v1.7.0
 	golang.org/x/sys v0.0.0-20211116061358-0a5406a5449c // indirect
-	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
 	google.golang.org/protobuf v1.27.1
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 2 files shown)


plan/sort.go
```diff
@@ -80,11 +80,27 @@ func Sort(p *blip.Plan) []SortedLevel {
 						Metrics: []string{},
 					}
 				}
-				higherDomain.Metrics = append(higherDomain.Metrics, lower.Collect[domain].Metrics...)
+
+				seen := toSet(higherDomain.Metrics)
```
daniel-nichter, line 84: When adding new, important blocks of logic, also add a code comment to explain what/why. Future engineers will thank you.

plan/sort_test.go
```diff
@@ -320,3 +320,310 @@ func TestSortComplex(t *testing.T) {
 		t.Error(diff)
 	}
 }
+
+func TestSort_PlanShouldBeUpdatedCorrectly(t *testing.T) {
```
daniel-nichter, line 324: Blip uses test name schema similar to Go: `TestSoemthing`, i.e. doesn't use underscores (are are pretty rare in Go in general). Also, all things should be updated/whatever correctly, so most of this test name is not actually descriptive/precise.

plan/sort_test.go
```diff
@@ -320,3 +320,310 @@ func TestSortComplex(t *testing.T) {
...
+	tests := []struct {
+		name     string
+		input    blip.Plan
+		expected blip.Plan
+	}{
+		{
+			name: "1 level",
+			input: blip.Plan{
+				Levels: map[string]blip.Level{
```
daniel-nichter, line 333: It seems existing tests cover most of these test cases already? If not, then what else are these tests covering?

plan/sort_test.go
```diff
@@ -320,3 +320,310 @@ func TestSortComplex(t *testing.T) {
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for i, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
```
daniel-nichter, line 623: `t.Run` isn't necessary here. Just exec the contents of this func directly in the test.

PR #35, opened 2022-05-12
plan/plan_loader: return deep copy of blip.Plan from Plan

allanliu:
The `blip.Plan` returned by `PlanLoader.Plan` was a shallow copy. Which introduces the behavior where duplicate metrics are created when `ChangePlan` is called more than once. Using a deep copy fixes this issue.

### DIFF
--- a/plan/plan_loader.go
+++ b/plan/plan_loader.go
@@ -4,8 +4,10 @@
 package plan
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
+	"encoding/gob"
 	"fmt"
 	"io/ioutil"
 	"os"
@@ -310,7 +312,7 @@ func (pl *Loader) Plan(monitorId string, planName string, db *sql.DB) (blip.Plan
 	}
 
 	blip.Debug("%s: loading plan %s from %s", monitorId, planName, pm.source)
-	return pm.plan, nil
+	return deepcopyPlan(&pm.plan)
 }
 
 func (pl *Loader) Print() {
@@ -550,4 +552,24 @@ func ValidatePlans(plans []blip.Plan) error {
 	}
 
 	return nil
+}
+
+func deepcopyPlan(p *blip.Plan) (blip.Plan, error) {
+	// Since the operation needed here is not performance critical we can piggy back off of gob
+	// encode/decode to make the deep copy and not rely on staying up-to-date with the blip.Plan's
+	// internal structure.
+	var (
+		b   bytes.Buffer
+		ret blip.Plan
+	)
+
+	if err := gob.NewEncoder(&b).Encode(p); err != nil {
+		return ret, err
+	}
+
+	if err := gob.NewDecoder(&b).Decode(&ret); err != nil {
+		return ret, err
+	}
+
+	return ret, nil
 }
--- a/plan/plan_loader.go
+++ b/plan/plan_loader.go
@@ -312,6 +312,8 @@ func (pl *Loader) Plan(monitorId string, planName string, db *sql.DB) (blip.Plan
 	}
 
 	blip.Debug("%s: loading plan %s from %s", monitorId, planName, pm.source)
+	// Since blip.Plan has field types that pass by reference (maps and slices), we want to the returned plan to
+	// be a deep copy to ensure the caller cannot modify the original shared plan.
 	return deepcopyPlan(&pm.plan)
 }
 
--- a/plan/plan_loader_test.go
+++ b/plan/plan_loader_test.go
@@ -1,8 +1,9 @@
 // Copyright 2022 Block, Inc.
 
-package plan
+package plan_test
 
 import (
+	"context"
 	"fmt"
 	"path/filepath"
 	"testing"
@@ -12,7 +13,11 @@ import (
 
 	"github.com/cashapp/blip"
 	//"github.com/cashapp/blip/dbconn"
+
+	"github.com/cashapp/blip/metrics"
+	"github.com/cashapp/blip/plan"
 	"github.com/cashapp/blip/proto"
+	"github.com/cashapp/blip/test/mock"
 	//	"github.com/cashapp/blip/test"
 )
 
@@ -25,7 +30,7 @@ const (
 func TestLoadDefault(t *testing.T) {
 	cfg := blip.DefaultConfig(false)
 
-	pl := NewLoader(nil)
+	pl := plan.NewLoader(nil)
 	if err := pl.LoadShared(cfg.Plans, nil); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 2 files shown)


plan/plan_loader.go
```diff
@@ -310,7 +312,9 @@ func (pl *Loader) Plan(monitorId string, planName string, db *sql.DB) (blip.Plan
 	}
 
 	blip.Debug("%s: loading plan %s from %s", monitorId, planName, pm.source)
-	return pm.plan, nil
+	// Since blip.Plan has field types that pass by reference (maps and slices), we want to the returned plan to
+	// be a deep copy to ensure the caller cannot modify the original shared plan.
```
daniel-nichter, line 316: Good comment. That'll surely save future engineer's day.

plan/plan_loader.go
```diff
@@ -551,3 +555,23 @@ func ValidatePlans(plans []blip.Plan) error {
 
 	return nil
 }
+
+func deepcopyPlan(p *blip.Plan) (blip.Plan, error) {
+	// Since the operation needed here is not performance critical we can piggy back off of gob
+	// encode/decode to make the deep copy and not rely on staying up-to-date with the blip.Plan's
+	// internal structure.
```
daniel-nichter, line 562: Also good comment: this isn't perf-critical and avoiding the hassle of struct changes is a good choice for this kind of approach.

plan/plan_loader_test.go
```diff
@@ -1,16 +1,17 @@
 // Copyright 2022 Block, Inc.
 
-package plan_test
+package plan
```
daniel-nichter, line 3: It's best to avoid this: `pkg_test` helps enforce clean design, clean interfaces/APIs.

What we really want to test is the actual code calls that indirectly cause this bug, so `ChangePan` iirc? Or maybe just something like `got := Sort(Sort(plan))`.

As this test stands now, it's really just testing `deepcopyPlan`, which is fine as an isolated unit test. But what we really want to know is: can we fetch a plan, sort it, then fetch same plan again, sort it again, and get a good 2nd plan?

plan/plan_loader_test.go
```diff
@@ -65,3 +66,65 @@ func TestLoadOneFile(t *testing.T) {
 	}
 	assert.Equal(t, gotPlans, expectPlans)
 }
+
+func TestPlan_ShouldReturnDeepCopyOfPlan(t *testing.T) {
```
daniel-nichter, line 70: Please don't use underscores in test names like this; it's not Blip convention (and, I'd argue, not common in Go--at least not the Go I've seen). It can be shorter (Go loves short), like `TestDeepCopy` (the "Plan" part is implied given this pkg is narrowly scoped around plans, so all tests relate to plans).

plan/plan_loader_test.go
```diff
@@ -65,3 +66,65 @@ func TestLoadOneFile(t *testing.T) {
...
+
+	expected := pl.sharedPlans[0].plan
+	assert.Equal(t, expected, got)
+	verifyDeepCopy(t, expected, got)
+}
+
+// verifyDeepCopy verifies is a is a deep copy of b by comparing address of the slices and maps in blip.Plan.
+// This solution is not ideal, but there really isn't a good way to compare reference object addresses in go.
+// The converse is to attempt to mutate the slices and maps embedded within blip.Plan, but even more verbose.
+func verifyDeepCopy(t *testing.T, a, b blip.Plan) {
```
daniel-nichter, line 116: I think it's best to avoid testing in non-test functions like this one because it makes test cases difficult to follow and modify. It's best when each `func TestX` case is truly isolated because then future engineers can understand properly functionality case by case (and move/change them independently, too).

plan/plan_loader_test.go
```diff
@@ -65,3 +66,65 @@ func TestLoadOneFile(t *testing.T) {
...
+
+	got, err := pl.Plan("", planName, nil)
+	require.NoError(t, err)
+
+	expected := pl.sharedPlans[0].plan
+	assert.Equal(t, expected, got)
+	verifyDeepCopy(t, expected, got)
+}
+
+// verifyDeepCopy verifies is a is a deep copy of b by comparing address of the slices and maps in blip.Plan.
```
daniel-nichter, line 113: Do we have to go that far: comparing addresses? Testing like this should both reproduce and verify the actual (right/wrong) output that motivated it: dupes. So we don't need to look at addresses; instead, we do what Blip did (get plan, sort, get plan again, sort again) then check: do we see this,
```
CommitLatency
CommitLatency
```
again (bug) or not (fixed)?

daniel-nichter (changes requested):

daniel-nichter (changes requested):
...

plan/plan_loader_test.go
allanliu: Gotcha. I agree that if it's not a convention in this code base, then I'll avoid it. But FWIW, it is quite common and part of examples listed in [the examples of the official godoc testing overview](https://pkg.go.dev/testing#hdr-Examples).

plan/plan_loader_test.go
allanliu: I definitely see the value in enforcing tests that are not dependent on understanding internal implementations of the objects at test. However, and for this specific package, it can be difficult to setup test conditions to properly test, and certainly difficult to keep the test code as concise. I looked at the code, and to test the `Plan` method, we have to load the plans via invoking `LoadShared` which also requires us to have knowledge of the internal logic of the methods and structures at play in order to setup the conditions to make `Plan` testable, not to mention quite a bit more boiler plate testing code (more things that we have to maintain in the future). Thoughts?

In regards to testing the function using `Sort` or `ChangePlan`. I can see value in those packages. But it certainly does not belong here as it creates unnecessary coupling across package boundaries.

IMO, unit test within specific packages need to be concise with that they are testing for and should not have knowledge on the specifics of what callers from the outside may use it for. The end behavior we want is the same, we want to return a deep copy of `Plan` because we don't want a caller to unintentionally modify the data `PlanLoader` is managing. Essentially the gist is the same as the above in that minimizing coupling across package boundaries will improve the maintainability of the code base.

plan/plan_loader_test.go
allanliu: Sure. I'll remove this abstraction.

plan/plan_loader_test.go
allanliu: Yeah I know 😞 . Unfortunately I could figure out a good way to test. The best ways I saw as verifying this issue was to either compare addresses (go does not give you a way to directly compare references) or to observe via side effects such as routines that would modify the `map` and `slice` fields of the structure. I don't feel strongly about having to compare addresses, so I'm more than open to verification that relies on attempting to modify the received plan and see if those changes are reflected original plan. However, I do strongly that we should not couple this with the source bug in the test for this package for the reasons noted in the comments above.

Since the topic of the verifying the exact bug in which this issue was surfaced, I will add a test for `ChangePlan` to capture this behavior.

Thoughts?

daniel-nichter (approved):

PR #39, opened 2022-06-07
Add event.trx collector

wen-hui-sun:
### Note
Add `event.trx` domain which contains `oldest` metric
Jira ticket: https://jira.sqprod.co/browse/SPIN-3539

### Testing
Tested against my local mysql
1. add config files:

blip.yaml
  ```
plans:
  files:
    - plan.yaml
monitors:
  - id: test1
    hostname: 127.0.0.1
    username: blip
    meta:
      repl-source: 127.0.0.1
  ```
  plan.yaml
  ```
---
performance:
  freq: 5s
  collect:
    event.trx:
      metrics:
        - oldest
  ```
2. run blip
```
cd ./blip/bin/blip
./blip
```
3. Blip print following line
```
blip 0.0.0
in 412.404µs: map[event.trx:[{Name:oldest Value:0 Type:2 Group:map[] Meta:map[]}]]
```

### DIFF
--- a/metrics/event.trx/trx.go
+++ b/metrics/event.trx/trx.go
@@ -7,15 +7,12 @@ import (
 	"database/sql"
 	"fmt"
 
-	"strconv"
-
 	"github.com/cashapp/blip"
-	"github.com/cashapp/blip/sqlutil"
 )
 
 const (
 	DOMAIN           = "event.trx"
-	OLDEST_TRX_QUERY = `SELECT UNIX_TIMESTAMP(NOW()) - UNIX_TIMESTAMP(MIN(trx_started)) AS time FROM information_schema.innodb_trx;`
+	OLDEST_TRX_QUERY = `SELECT COALESCE(UNIX_TIMESTAMP(NOW()) - UNIX_TIMESTAMP(MIN(trx_started)), 0) t FROM information_schema.innodb_trx;`
 )
 
 type trxMetrics struct {
@@ -49,7 +46,7 @@ func (c *Trx) Domain() string {
 func (c *Trx) Help() blip.CollectorHelp {
 	return blip.CollectorHelp{
 		Domain:      DOMAIN,
-		Description: "Event transaction metrics from information_schema.innodb_trx",
+		Description: "Transaction metrics",
 		Options:     map[string]blip.CollectorHelpOption{},
 		Metrics: []blip.CollectorMetric{
 			{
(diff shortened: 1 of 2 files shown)


metrics/event.trx/trx.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"strconv"
+
+	"github.com/cashapp/blip"
+	"github.com/cashapp/blip/sqlutil"
+)
+
+const (
+	DOMAIN           = "event.trx"
+	OLDEST_TRX_QUERY = `SELECT UNIX_TIMESTAMP(NOW()) - UNIX_TIMESTAMP(MIN(trx_started)) AS time FROM information_schema.innodb_trx;`
```
daniel-nichter, line 18: Can simplify and fix `NULL` handling (when there are no active trx):
```sql
SELECT COALESCE(NOW() - MIN(trx_started), 0) t FROM information_schema.innodb_trx
```

metrics/event.trx/trx.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Collect collects metrics at the given level.
+func (c *Trx) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	rm, ok := c.atLevel[levelName]
+	if !ok {
+		return nil, nil
+	}
+
+	res, err := sqlutil.RowToMap(ctx, c.db, OLDEST_TRX_QUERY)
```
daniel-nichter, line 100: Given the aggregate func in the query, you don't need `RowToMap`: there'll only be 1 row. Instead, use [QueryRowContext](https://pkg.go.dev/database/sql#DB.QueryRowContext).

metrics/event.trx/trx.go
```diff
@@ -0,0 +1,122 @@
...
+// Domain returns the Blip metric domain name (DOMAIN const).
+func (c *Trx) Domain() string {
+	return DOMAIN
+}
+
+// Help returns the output for blip --print-domains.
+func (c *Trx) Help() blip.CollectorHelp {
+	return blip.CollectorHelp{
+		Domain:      DOMAIN,
+		Description: "Event transaction metrics from information_schema.innodb_trx",
```
daniel-nichter, line 52: Let's go with just "Transaction metrics"  because trx are a type of even (they're not "event transactions" per-se) and Blip tries to hide implementation details like information_schema.innodb_trx so they can vary (as we discussed, it'd technically be nicer to use perf_schema for this).

daniel-nichter (changes requested):

metrics/event.trx/trx.go
wen-hui-sun: Done

metrics/event.trx/trx.go
wen-hui-sun: Good callout! Updated

daniel-nichter (approved):

PR #44, opened 2022-07-14
Added event.stmt collector

JashLal:
Implemented a collector for the event.stmt domain with two metrics...

- oldestQuery
- activeLongRunningQueries


metrics/stmt/metric.go
```diff
@@ -0,0 +1,67 @@
...
+	metric
+}
+
+func (q activeLongQueryCount) CollectMetric(ctx context.Context, db *sql.DB) (float64, error) {
+	var t float64
+
+	query := `SELECT COUNT(*) FROM performance_schema.events_statements_current 
+		WHERE END_EVENT_ID IS NULL 
+		AND EVENT_NAME NOT IN ('statement/com/Sleep','statement/com/Connect','statement/com/Binlog Dump','statement/com/Binlog Dump GTID') 
+		AND TIMER_WAIT > 30000000000000;`
```
daniel-nichter, line 59: What's the magic value, `30000000000000`?

metrics/stmt/metric.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+type oldestQuery struct {
+	metric
+}
+
+func (q oldestQuery) CollectMetric(ctx context.Context, db *sql.DB) (float64, error) {
+	var t float64
+
+	query := `SELECT TIMER_WAIT FROM performance_schema.events_statements_current
```
daniel-nichter, line 33: https://hackmysql.com/golang/go-antipatterns/#non-variable-variable

This one and the other one, too.

metrics/stmt/metric.go
```diff
@@ -0,0 +1,67 @@
...
+	metric
+}
+
+func (q oldestQuery) CollectMetric(ctx context.Context, db *sql.DB) (float64, error) {
+	var t float64
+
+	query := `SELECT TIMER_WAIT FROM performance_schema.events_statements_current 
+		WHERE END_EVENT_ID IS NULL 
+		AND EVENT_NAME NOT IN ('statement/com/Sleep','statement/com/Connect','statement/com/Binlog Dump','statement/com/Binlog Dump GTID') 
+		ORDER BY TIMER_WAIT DESC LIMIT 1;`
```
daniel-nichter, line 36: Nice: you didn't try to `MAX()` on an invalid full group by. :-)

metrics/stmt/metric.go
JashLal: It's 30 seconds in picoseconds

metrics/stmt/metric.go
daniel-nichter: Go with proud: I've seen it many times before; yours was the one that put it on the board, so to speak.

metrics/stmt.current/current.go
```diff
@@ -0,0 +1,168 @@
...
+// Help returns the output for blip --print-domains.
+func (c *Current) Help() blip.CollectorHelp {
+	return blip.CollectorHelp{
+		Domain:      DOMAIN,
+		Description: "Statement metrics",
+		Options: map[string]blip.CollectorHelpOption{
+			OPT_THRESHOLD: {
+				Name:    OPT_THRESHOLD,
+				Desc:    "The length of time (in microseconds) that a query must be active to be considered slow",
+				Default: "30000000",
```
daniel-nichter, line 57: Let's use a duration string and default "1.0s" and minimum value 1 us. Reason being: I was trying to think of a reasonable default (30s is way to long imho)... but we'll never know what users think is "long". So 1.0 is a nice "round" starting point (1s queries _are_ pretty slow in a normal-ish MySQL db, I'd say). Then since we don't know what users will want, the millisecond format is too messy; duration strings are easy to write for any range of values.

metrics/stmt.current/current.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+
+		threshold, ok := dom.Options[OPT_THRESHOLD]
+		if !ok {
+			threshold = c.Help().Options[OPT_THRESHOLD].Default
+		}
+		t, err := strconv.ParseFloat(threshold, 64)
+		if err != nil {
+			t = 0
```
daniel-nichter, line 105: Return the error. (This will change to parsing a time duration string.)

metrics/stmt.current/current.go
```diff
@@ -0,0 +1,168 @@
...
+
+	"github.com/cashapp/blip"
+)
+
+const (
+	DOMAIN = "stmt.current"
+	query  = `SELECT TIMER_WAIT FROM performance_schema.events_statements_current 
+		WHERE END_EVENT_ID IS NULL 
+		AND EVENT_NAME NOT LIKE ('statement/com/Binlog%')`
+	OPT_THRESHOLD = "threshold"
```
daniel-nichter, line 18: Let's call this option `slow-threshold` to give a little more context (because just "threshold" in the stmt.current domain could be a threshold for anything).

daniel-nichter (approved):

PR #46, opened 2022-07-20
Pass TransformMetrics down to LevelCollector

wen-hui-sun:
### Note
TransformMetrics is not set in LevelCollector so it cannot be invoked to convert metrics even it is specified by the client. This change is to pass TransformMetrics down to LevelCollector. 

### Testing
Pointed to my local blip change and tested against test metacluster node with TransformMetrics configuration


daniel-nichter (approved):

daniel-nichter:
I forgot to connect the wires. :-(  Thanks for fix this!

PR #47, opened 2022-07-20
[SPIN-3566] adding table size collector

serjSquare:


metrics/size.table/table.go
```diff
@@ -0,0 +1,131 @@
...
+const (
+	DOMAIN = "size.table"
+
+	OPT_ALL = "all"
+
+	TABLE_SIZES_QUERY = `
+    SELECT table_schema AS db, table_name as tbl,
+           data_length + index_length AS tbl_size_bytes
+      FROM information_schema.TABLES
+     WHERE table_schema NOT IN ('performance_schema', 'information_schema', 'mysql');`
```
daniel-nichter, line 22: Make this an option (`OPT_SCHEMA_FILTER` perhaps?) with default value equal to this + `sys`.

metrics/size.table/table.go
```diff
@@ -0,0 +1,131 @@
...
+		t.query[level.Name] = TABLE_SIZES_QUERY
+		if dom.Options[OPT_ALL] == "yes" {
+			t.total[level.Name] = true
+		}
+	}
+	return nil, nil
+}
+
+func (t *Table) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	if !t.total[levelName] {
```
daniel-nichter, line 85: Use `query` to check. Total is optional per level.

metrics/size.table/table.go
```diff
@@ -0,0 +1,131 @@
...
+		var ok bool
+		m.Value, ok = sqlutil.Float64(val)
+		if !ok {
+			continue
+		}
+		total += m.Value
+		metrics = append(metrics, m)
+	}
+
+	metrics = append(metrics, blip.MetricValue{
```
daniel-nichter, line 123: This block should be optional: only if total=true.

metrics/size.table/table.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	return nil, nil
+}
+
+func (t *Table) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	if !t.total[levelName] {
+		return nil, nil
+	}
+
+	rows, err := t.db.QueryContext(ctx, TABLE_SIZES_QUERY)
```
daniel-nichter, line 89: `s/TABLE_SIZES_QUERY/value from query map/`

metrics/size.table/table.go
```diff
@@ -0,0 +1,157 @@
...
+				Default: "yes",
+				Values: map[string]string{
+					"yes": "Includes total size of all tables",
+					"no":  "Excludes total size of all tables",
+				},
+			},
+			OPT_SCHEMA_FILTER: {
+				Name:    OPT_SCHEMA_FILTER,
+				Desc:    "Excludes performance_schema, information_schema, mysql, and sys",
+				Default: "yes",
```
daniel-nichter, line 63: Default "performance_schema, information_schema, mysql, and sys" (see `size.database`), so if user sets option they totally override default and set their own filter. E.g.if they want to do just +foo, then they'll have to do
```yaml
exclude: "performance_schema, information_schema, mysql, sys, foo"
```

But let's think about this more holistically because `size.table` is a little more complicated than `size.database`: ideally, we want to allow user to filter by schema _and/or_ table. E.g. what if they just want to exclude one specific table? 🤔 

Let's do similar to `size.database` but slightly different. So the options are:

```
	OPT_INCLUDE = "include"
	OPT_EXCLUDE = "exclude"
```

But the values can be:

* `foo` = any table named food (in any db)
* `foo.*` = all tables in db `foo`
* `foo.bar` = table `bar` in db `foo`

I think that makes sense in this context, and `foo.*` is  a MySQL thing (used for user grants). See what others on ODS think.

So really we'll have option `exclude` default to `performance_schema.*, information_schema.*, mysql.*, and sys.*`.

metrics/size.table/query.go
```diff
@@ -0,0 +1,19 @@
...
+func TableSizeQuery(set map[string]string, def blip.CollectorHelp) (string, error) {
+	query := `
+	SELECT table_schema AS db, table_name as tbl,
+		COALESCE(data_length + index_length, 0) AS tbl_size_bytes
+		FROM information_schema.TABLES`
+
+	if val := set[OPT_SCHEMA_FILTER]; val == "no" {
+		query = query + ` WHERE table_schema NOT IN ('performance_schema', 'information_schema', 'mysql', 'sys')`
+	}
+	query = query + `;`
```
daniel-nichter, line 16: You don't need terminating `;` in code (unless you're doing multi-statements like `query := "foo; bar".  The `;` is an interactive shell thing, i.e. when humans input queries in the `mysql` shell.

metrics/size.table/table.go
```diff
@@ -0,0 +1,157 @@
...
+	total := float64(0)
+	for rows.Next() {
+		if err = rows.Scan(&dbName, &tblName, &val); err != nil {
+			return nil, err
+		}
+
+		m := blip.MetricValue{
+			Name:  "bytes",
+			Type:  blip.GAUGE,
+			Group: map[string]string{"db": dbName, "table": tblName},
```
daniel-nichter, line 136: `s/table/tbl/`  I see now that I forgot to document that Blip prefers the shortest possible name/term that's not too cryptic because it generates a lot of data (especially over time), so every byte counts.

metrics/size.table/table.go
```diff
@@ -0,0 +1,166 @@
...
+
+		q, err := TableSizeQuery(dom.Options)
+		if err != nil {
+			return nil, err
+		}
+		t.query[level.Name] = q
+
+		if dom.Options[opt_total] == "yes" {
+			t.total[level.Name] = true
+		} else {
```
daniel-nichter, line 109: You don't need this `else {` branch: `false` is the zero value for bools. (And Go commonly relies on its zero values for such things.)

metrics/size.table/query.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"fmt"
+	"strings"
+)
+
+func TableSizeQuery(set map[string]string) (string, error) {
+	query := "SELECT table_schema AS db, table_name as tbl, COALESCE(data_length + index_length, 0) AS tbl_size_bytes FROM information_schema.TABLES"
+	where := " WHERE "
+	if include := set[OPT_INCLUDE]; include != "" {
+		var include_tables []string
```
daniel-nichter, line 12: Go style is camelCase, not snake_case (except, sometimes, for const and test func names).

metrics/size.table/table.go
```diff
@@ -0,0 +1,166 @@
...
+	"github.com/cashapp/blip/sqlutil"
+)
+
+const (
+	DOMAIN = "size.table"
+
+	opt_total         = "total"
+	OPT_EXCLUDE       = "exclude"
+	OPT_INCLUDE       = "include"
+	OPT_SCHEMA_FILTER = "schema"
```
daniel-nichter, line 19: Dead code.

metrics/size.table/query.go
```diff
@@ -0,0 +1,55 @@
...
+		}
+		if i != (len(tables) - 1) {
+			if isInclude {
+				where = where + " OR "
+			} else {
+				where = where + " AND NOT "
+			}
+		}
+	}
+	return where
```
daniel-nichter, line 54: Where is this var declared? It seems this func is returning one of its modify args, which is not usual.

metrics/size.table/query.go
```diff
@@ -0,0 +1,55 @@
...
+		if strings.Contains(set[OPT_INCLUDE], ",") {
+			include_tables = strings.Split(set[OPT_INCLUDE], ",")
+		} else {
+			include_tables = []string{set[OPT_INCLUDE]}
+		}
+		where = setWhere(include_tables, true, where)
+	} else {
+		where = where + "NOT "
+		var exclude_tables []string
+		if strings.Contains(set[OPT_EXCLUDE], ",") {
```
daniel-nichter, line 22: More idiomatic
```go
excludeTables := strings.Split(set[OPT_EXCLUDE], ",")
if len(excludeTables) > 0 {
    setWhere(...)
}
```

Or more compactly:

```go
setWhere(strings.Split(set[OPT_EXCLUDE], ","), <other args>)
```
because a `for` over an empty list (in `setWhere()`) will simply do nothing.

Either way, the `else {` branch here isn't needed or the standard way of writing Go.

metrics/size.table/query.go
```diff
@@ -0,0 +1,55 @@
...
+				where = where + fmt.Sprintf("(table_schema = '%s' AND table_name = '%s')", db, table)
+			}
+		} else {
+			where = where + fmt.Sprintf("(table_name = '%s')", exclude_table)
+		}
+		if i != (len(tables) - 1) {
+			if isInclude {
+				where = where + " OR "
+			} else {
+				where = where + " AND NOT "
```
daniel-nichter, line 50: Why,
```
SELECT table_schema AS db, table_name as tbl, COALESCE(data_length + index_length, 0) AS tbl_size_bytes FROM information_schema.TABLES WHERE NOT (table_schema = 'mysql') AND NOT (table_schema = 'information_schema') AND NOT (table_schema = 'performance_schema') AND NOT (table_schema = 'sys');
```

and not more simply,

```
SELECT table_schema AS db, table_name as tbl, COALESCE(data_length + index_length, 0) AS tbl_size_bytes FROM information_schema.TABLES WHERE table_schema NOT IN ('mysql','information_schema','performance_schema','sys');
```
? 

Granted, you'd still have to handle cases for `table_name [NOT] IN (...)`, too.  Point is: `IN()` is more compact and more common SQL.

metrics/size.table/query_test.go
```diff
@@ -0,0 +1,39 @@
...
+package sizetable_test
+
+import (
+	"fmt"
+	"testing"
+
+	sizetable "github.com/cashapp/blip/metrics/size.table"
+)
+
+func TestTableSizeQuery(t *testing.T) {
```
daniel-nichter, line 12: Thank you for testing. :-)

metrics/size.table/table.go
```diff
@@ -0,0 +1,166 @@
...
+		} else {
+			t.total[level.Name] = false
+		}
+	}
+	return nil, nil
+}
+
+func (t *Table) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	q, ok := t.query[levelName]
+
```
daniel-nichter, line 118: Don't have blank lines between this kind of Go boilerplate. The `ok` var belongs right next to where it's created.

metrics/size.table/table.go
```diff
@@ -0,0 +1,166 @@
...
+	if !ok {
+		return nil, nil
+	}
+	rows, err := t.db.QueryContext(ctx, q)
```
daniel-nichter, line 122: But do add a blank line above here because now you're separating different code blocks. At this point, we're ok (`ok=true`), so next code block: do the query, check the error (boilerplate), and defer the rows--a whole block of Go boilerplate code for this type of work.

metrics/size.table/query.go
serjSquare: Good call. I didn't really think of it that way. I'll change it so it isn't editing and returning the argument which is passed to it.

metrics/size.table/query_test.go
serjSquare: Anytime 👍

metrics/size.table/query.go
serjSquare: I thought that if I try to split a string on something that it didn't contain would result in an error, but I realize now after trying it out in golang playground that it doesn't.

metrics/size.table/query.go
serjSquare: if we do:
`table_schema NOT IN ('my_sql', 'sys') and table_name NOT IN ('table_1', 'table_2')`

My worry is that it'll include/exclude tables and dbs which we don't intend on targeting. If our 'my_sql' and 'sys' dbs looked like:

sys: table_1
my_sql: table_1, table_2

and we entered an input like this to exclude table_1 from my_sql and table_2 from sys:
`my_sql.table_1,sys.table_2`

Using this formatting:
- foo = any table named food (in any db)
- foo.* = all tables in db foo
- foo.bar = table bar in db foo

the query would exclude table_1 from the sys db as well even when that wasn't the users intention.

daniel-nichter: Right... my example was not a good one, but it's ok: we can start with this method since it works and table size collection (should be) low frequency, so performance isn't as critical.

metrics/size.table/table.go
daniel-nichter: I don't see where `OPT_SCHEMA_FILTER` is used?

daniel-nichter (approved):

daniel-nichter:
Apart from `const OPT_SCHEMA_FILTER`, looks good.

metrics/size.table/table.go
serjSquare: sorry I don't think I pushed the changes

PR #49, opened 2022-07-21
Prudhvi.tls fix

prudhvi:
Handling the case correctly only if CA for TLS is given but not cert and key.
Logging error if TLS can't be enabled successfully.


test/mycnf/full-dsn
```diff
@@ -3,6 +3,5 @@ user     = U
 password = P
 host     = H
 port     = 33560
-ssl-ca   = CA
-ssl-cert = CERT
-ssl-key  = KEY
+# Path is relative to the test file factory_test.go
+ssl-ca   = ../test/mycnf/rds-master.pem
```
prudhvi, line 7: Would be nice to get this to just `./rds-master.pem` but currently the path should be relative to the script that's using this full-dsn file.

dbconn/factory.go
```diff
@@ -111,7 +111,13 @@ func (f factory) Make(cfg blip.ConfigMonitor) (*sql.DB, string, error) {
 
 	// Load and register TLS, if any
 	tlsConfig, err := cfg.TLS.LoadTLS()
-	if tlsConfig != nil && err != nil {
+
+	if err != nil {
+		blip.Debug("Error loading TLS Config for %s", cfg.MonitorId)
```
daniel-nichter, line 116: Classic error message mistake: an error that doesn't say what the error is :-)  Either add `err` to the debug, or if the caller debug's the error line, then this Debug is superfluous

daniel-nichter (approved):

dbconn/factory.go
prudhvi: Yeah thought about it,  i will be printing the error twice, so didn't print it here , caller will be logging the returned error anyway.

prudhvi: Cleaned up by removing the whole debug, also I was looking for patterns to say logger.Error or logger.Warn but I only see them available in few file(main.go or blip.go) .Is that intentional? Most other files only have blip.Debug available.

PR #53, opened 2022-07-25
Add option to skip TLS verification

prudhvi:


prudhvi:
Still working, had to handle the case where nothing is supplied under TLS

config.go
```diff
@@ -959,12 +961,19 @@ func (c *ConfigTLS) InterpolateMonitor(m *ConfigMonitor) {
 }
 
 func (c ConfigTLS) LoadTLS() (*tls.Config, error) {
+
+	tlsConfig := &tls.Config{}
+
+	// Skip TLS verification
+	if c.SkipVerify != nil && *c.SkipVerify {
```
daniel-nichter, line 968: `blip.True()`

daniel-nichter: Move this block under `if c.Cert == "" && c.Key == "" && c.CA == "" {`.  This func is supposed to return `nil, nil` if there's no TLS config.  So skip-verify logic needs to happen only if there is _some_ TLS config.

prudhvi: so you mean
```
tls:
  skip-verify: true
  ``` 
  is invalid because they CA or {cert,key} isn't supplied?
  
  So a valid skip verify config is only one of the below?
```
tls:
  skip-verify: true
  ca: path_to_ca
  ```
or 
```
tls:
    skip-verify: true
     cert: path_to_cert
     key: path_to_key
     ```

Comment (daniel-nichter) [reply config.go]: Yes, that's correct.

Comment (prudhvi) [reply config.go]: Updated, but any strong reason why we mandate those even if we are asking server to ignore certs/CA anyway?

File: config.go [hunk 2]
```diff
@@ -965,6 +968,12 @@ func (c ConfigTLS) LoadTLS() (*tls.Config, error) {
 
 	tlsConfig := &tls.Config{}
 
+	// Skip TLS verification
+	if True(c.SkipVerify) {
+		tlsConfig.InsecureSkipVerify = true
+		return tlsConfig, nil
```
daniel-nichter, line 974: Don't return because returning here makes this func ignore all its logic (the code that follows it). Skip-verify is just a flag _to_ the config; it's not the entirety of the config.

PR #57, opened 2022-07-28
Add query.global collector

wen-hui-sun:
### Note
Add  query.global collector for response time metric

### Testing
Tested against my test metacluster node


metrics/percona/qrt.go
```diff
@@ -226,39 +225,12 @@ func (c *QRT) prepareLevel(dom blip.Domain, level blip.Level) error {
 	percentilesList := strings.Split(strings.TrimSpace(percentilesStr), ",")
 
 	for _, percentileStr := range percentilesList {
-		percentileStr = strings.TrimSpace(percentileStr)
-		f, err := strconv.ParseFloat(percentileStr, 64)
+		percentile, err := util.ParsePercentileStr(percentileStr)
```
daniel-nichter, line 228: Nice refactoring: common code, short clear functions, unit tested 💯

metrics/query.global/global.go
```diff
@@ -0,0 +1,194 @@
...
+	TRUNCATE_QUERY = "TRUNCATE TABLE performance_schema.events_statements_histogram_global"
+)
+
+type Global struct {
+	db *sql.DB
+	// --
+	optional      map[string]bool      // keyed on level
+	percentiles   map[string][]float64 // keyed on level
+	truncateTable map[string]bool      // keyed on level
+	available     bool
```
daniel-nichter, line 33: `s/available/enabled/`  Perf Schema is always available; question is whether or not it's enabled. (Fwiw, it's enabled by default as of 5.7 iirc, but Aurora _disables_ it by default 🤦 )

metrics/query.global/global.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return metrics, nil
+}
+
+// prepareLevel sanitizes options for given level based on user specified options
+func (c *Global) prepareLevel(dom blip.Domain, level blip.Level) error {
```
daniel-nichter, line 161: Let's move this into `Prepare` because it is the heart of that func. Factoring it out (in this case) adds no value; it just splits up a func's main code, which leads the reader around needlessly.

metrics/query.global/global.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	return nil, nil
+}
+
+// Collect collects metrics at the given level.
+func (c *Global) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	if !c.available {
```
daniel-nichter, line 119: I think you can just `return nil, nil` if not available. I need to double check but I'm pretty sure I set a bad example by checking `!c.optional[levelName]` in `Collect`, and now other devs have followed my bad example. The caller (`Monitor.Engine`) should only call `Collect` if and when it's correct to do so.

metrics/query.global/global.go
```diff
@@ -0,0 +1,194 @@
...
+			panic(fmt.Sprintf("%s: required response time metrics couldn't be collected", levelName))
+		} else {
+			return nil, nil
+		}
+	}
+
+	var metrics []blip.MetricValue
+	for _, percentile := range c.percentiles[levelName] {
+		where := fmt.Sprintf(" WHERE bucket_quantile >= %f ORDER BY bucket_quantile LIMIT 1", percentile)
+		query := BASE_QUERY + where
```
daniel-nichter, line 130: Do all preparation only in `Prepare`. So you probably want to save something like,
```
map[string]string{
    "p999": "<query for p999>"
}```

File: metrics/query.global/global.go [hunk 6]
```diff
@@ -0,0 +1,194 @@
...
+
+// Help returns the output for blip --print-domains.
+func (c *Global) Help() blip.CollectorHelp {
+	return blip.CollectorHelp{
+		Domain:      DOMAIN,
+		Description: "Collect metrics for all queries",
+		Options: map[string]blip.CollectorHelpOption{
+			OPT_RESPONSE_TIME_OPTIONAL: {
+				Name:    OPT_RESPONSE_TIME_OPTIONAL,
+				Desc:    "If collecting response time metrics is optional. This will fail if it is required but not available",
```
daniel-nichter, line 61: I need to think more about this kind of option (here and in other collectors). We want something like a 3-part option:
* Required (error in Prepare if not)
* Optional (try but no error on failure)
Granted, we kind of have that now... I'm just thinking "optional" is not the best option name for this 🤔

metrics/query.global/global.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/cashapp/blip/metrics/util"
+)
+
+const (
+	DOMAIN = "query.global"
+
+	OPT_RESPONSE_TIME_PERCENTILES = "response-time-percentiles"
+	OPT_RESPONSE_TIME_OPTIONAL    = "response-time-optional"
+	OPT_TRUNCATE_TABLE            = "truncate-table"
+	DEFAULT_PERCENTILE_OPTION     = "999"
```
daniel-nichter, line 21: @joycse06 and I talked about this, and seeing it in action again, "999" might have been a bad idea on my part. It's implicit for p999, but maybe _too_ implicit. When writing about such things, no one would write just "999"; it has to be "p999" to convey the fact that it's a percentile. The same is true for "99" and "99.9" but those are a little more clear since percentiles must range `[0, 100]`, which is why "999" is odd since it's out of range. Perhaps we make this a little more strict (i.e. the sqlutil code you factored out) and support: "0" through "100" (float) and `^[pP]\d+/$` ?

daniel-nichter: Thought about this more and, if you could please, let's change this to "0" through "100" (float) and `/^[pP]\d+$/`.  That's more consistent with established norms (e.g. "p99" is a norm, and "99" is at least within a logical range: `[0, 100]`.)

wen-hui-sun: yes, I can refactor

metrics/query.global/global.go
wen-hui-sun: `Optional` is kind of confusing at the first glance. Maybe this option can be something more explicit like `ignore-unavailable-error`.

metrics/query.global/global.go
wen-hui-sun: The check here depends on the options specified by clients. If the metric is required but not available, we fail the collector with panic. Maybe we can do such check in `Prepare`?

metrics/query.global/global.go
wen-hui-sun: Good call out. Will do

daniel-nichter (approved):

PR #60, opened 2022-08-16
Run spellcheck on docs

morgo:
This runs a spell check on the docs (using [vscode spellchecker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker)).

I've manually reviewed the spell suggestions, so there should not be false positives.

### DIFF
--- a/docs/v1.0/config/interpolation.md
+++ b/docs/v1.0/config/interpolation.md
@@ -46,7 +46,7 @@ monitors:
       hostname: db.local
 ```
 
-The result is `monitors.hostname = "db.local"` because `%{monitor.tags.hostname}` refers to the local `monitrs.tags.hostname` variable.
+The result is `monitors.hostname = "db.local"` because `%{monitor.tags.hostname}` refers to the local `monitors.tags.hostname` variable.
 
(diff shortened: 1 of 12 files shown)


docs/v1.0/config/interpolation.md
```diff
@@ -46,7 +46,7 @@ monitors:
       hostname: db.local
 ```
 
-The result is `monitors.hostname = "db.local"` because `%{monitor.tags.hostname}` refers to the local `monitrs.tags.hostname` variable.
+The result is `monitors.hostname = "db.local"` because `%{monitor.tags.hostname}` refers to the local `monitors.tags.hostname` variable.
```
morgo, line 49: Just wanted to check this is correct - monitors.tags.X vs monitor.tags.hostname?

daniel-nichter: Yes, correct enough because haven't settled on a convention yet for this ("monitor" or "monitors": probably the latter since it's `monitors:` in the config file).

daniel-nichter (approved):

PR #62, opened 2022-08-19
Refactor percona qrt

wen-hui-sun:
### Note
Refactor percona qrt to make it consistent with query.response-time

### Testing
Tested against my test cluster node

### DIFF
--- a/metrics/percona/qrt.go
+++ b/metrics/percona/qrt.go
@@ -8,7 +8,10 @@ import (
 	"fmt"
 	"strings"
 
+	myerr "github.com/go-mysql/errors"
+
 	"github.com/cashapp/blip"
+	"github.com/cashapp/blip/errors"
 	"github.com/cashapp/blip/sqlutil"
 )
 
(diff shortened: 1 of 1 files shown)


metrics/percona/qrt.go
```diff
@@ -39,31 +42,40 @@ const (
 
 const (
 	OPT_PERCENTILES           = "percentiles"
-	OPT_OPTIONAL              = "optional"
+	OPT_REAL_PERCENTILES      = "real-percentiles"
 	OPT_FLUSH_QRT             = "flush"
 	default_percentile_option = "999"
+
+	ERR_UNKNOWN_TABLE = "table-is-unknown"
```
daniel-nichter, line 49: Weird... for P_S. events_statements_histogram_global MySQL returns error 1146, but error 1109 for this one. In any case, let's call this `"unknown-table"` to match the const name and MySQL error message.

daniel-nichter (approved):

metrics/percona/qrt.go
wen-hui-sun: Yeah, the error is different. 
Updated to `unknown-table`

PR #63, opened 2022-08-25
Copy options and errors in sort

wen-hui-sun:
### Note
Copy options and errors in sort when the frequency of a level is a multiple of another level

### Testing
Tested locally

### DIFF
--- a/plan/sort.go
+++ b/plan/sort.go
@@ -84,10 +84,14 @@ func Sort(p *blip.Plan) []SortedLevel {
 				}
 				higherDomain.Metrics = append(higherDomain.Metrics, lower.Collect[domain].Metrics...)
 				for k, v := range lower.Collect[domain].Options {
-					higherDomain.Options[k] = v
+					if _, ok := higherDomain.Options[k]; !ok {
+						higherDomain.Options[k] = v
+					}
 				}
 				for k, v := range lower.Collect[domain].Errors {
-					higherDomain.Errors[k] = v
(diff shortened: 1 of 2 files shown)


plan/sort.go
```diff
@@ -78,9 +78,17 @@ func Sort(p *blip.Plan) []SortedLevel {
 					higherDomain = blip.Domain{
 						Name:    domain,
 						Metrics: []string{},
+						Options: map[string]string{},
+						Errors:  map[string]string{},
 					}
 				}
 				higherDomain.Metrics = append(higherDomain.Metrics, lower.Collect[domain].Metrics...)
+				for k, v := range lower.Collect[domain].Options {
+					higherDomain.Options[k] = v
```
daniel-nichter, line 87: This is correct in general, but I think we also need to _not_ overwrite options set in higher domain. So a pathological case would be:

```
level_1:
  freq: 5s
  collect:
    foo:
      options:
        x: aaa
level_2:
  freq: 10s
  collect:
    foo:
      options:
        x: bbb
```

Option `x=aaa` should not overwrite explicitly set option `x=bbb` in the higher level. If I'm thinking clearly about this and you agree, let's also write a new unit test case for this.

daniel-nichter (approved):

plan/sort.go
wen-hui-sun: Talked offline. It makes sense to pick the options and errors of higher level.

daniel-nichter (approved):

PR #64, opened 2022-09-09
Setup sink client timeout and close idle connections for sfx

wen-hui-sun:
### Note
Setup http client timeout when creating sink client and close idle connection after a sfx error to fix no datapoints issue in sfx for ~6min.

### Testing
Tested against test cluster nodes

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -52,7 +52,9 @@ type httpClientFactory struct {
 }
 
 func (f httpClientFactory) MakeForSink(sinkName, monitorId string, opts, tags map[string]string) (*http.Client, error) {
-	client := &http.Client{}
+	client := &http.Client{
+		Timeout: 5 * time.Second,
+	}
 	if f.cfg.Proxy != "" {
 		proxyFunc := func(req *http.Request) (url *url.URL, err error) {
 			return url.Parse(f.cfg.Proxy)
--- a/server/server.go
+++ b/server/server.go
@@ -52,9 +52,7 @@ type httpClientFactory struct {
 }
 
 func (f httpClientFactory) MakeForSink(sinkName, monitorId string, opts, tags map[string]string) (*http.Client, error) {
-	client := &http.Client{
-		Timeout: 5 * time.Second,
-	}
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -52,7 +52,9 @@ type httpClientFactory struct {
 }
 
 func (f httpClientFactory) MakeForSink(sinkName, monitorId string, opts, tags map[string]string) (*http.Client, error) {
-	client := &http.Client{}
+	client := &http.Client{
+		Timeout: 5 * time.Second,
```
daniel-nichter, line 56: We shouldn't hard-code this because we can't know what's valid for all users. Instead, you'll want to implement https://github.com/cashapp/blip/blob/main/blip.go#L108

sink/signalfx.go
```diff
@@ -173,7 +173,13 @@ func (s *SignalFx) Send(ctx context.Context, m *blip.Metrics) error {
 
 	// Send metrics to SFX. The SFX client handles everything; we just pass
 	// it data points.
-	return s.sfxSink.AddDatapoints(ctx, dp[0:n])
+	err := s.sfxSink.AddDatapoints(ctx, dp[0:n])
+	if err != nil {
+		blip.Debug("error sending data points to SignalFx: %s", err)
+		s.sfxSink.Client.CloseIdleConnections()
```
daniel-nichter, line 179: I've not seen this used before. I'm not sure that an error means we need to close idle conns, but I also don't see how it'd hurt. Maybe too many timeouts create too many conns, which become idle?

server/server.go
wen-hui-sun: Good catch. It makes sense to implement client specific httpclient.

sink/signalfx.go
wen-hui-sun: I tried testing without close idle connections and it turned out sfx had no data points for 10+ min due to errors. Our old monitoring has such fix as well for this similar issue.

daniel-nichter (approved):

PR #67, opened 2022-09-26
Adding DataDog Sink

Areson:
Adding a sink for DataDog. Updating `MetricValue` to include a `Frequency` field which records the frequency the metric is collected at. This value is need by DataDog's `COUNT` metric type in order to display metrics properly in the UI at various resolutions.

Added `collectorFrequencyWrapper` type to wrap collectors and update the `Frequency` field on the returned `MetricValue`s.

### DIFF
--- a/blip.go
+++ b/blip.go
@@ -56,7 +56,7 @@ type MetricValue struct {
 	// Boolean values are reported as 0 and 1.
 	Value float64
 
-	// Type is the metric type: COUNTER, COUNTER, and other const.
+	// Type is the metric type: GAUGE, COUNTER, and other const.
 	Type byte
 
 	// Group is the set of name-value pairs that determine the group to which
@@ -65,6 +65,10 @@ type MetricValue struct {
 
 	// Meta is optional key-value pairs that annotate or describe the metric value.
 	Meta map[string]string
+
+	// The frequency that this metric is collected at. Some sinks need this value
+	// to properly process COUNTER type metrics.
+	Frequency int
 }
 
 // Sink sends metrics to an external destination.
--- a/metrics/factory.go
+++ b/metrics/factory.go
@@ -75,7 +75,7 @@ func Exists(domain string) bool {
 // Make makes a metric collector for the domain using a previously registered factory.
 //
 // See types in the blip package for more details.
-func Make(domain string, args blip.CollectorFactoryArgs) (blip.Collector, error) {
+func Make(domain string, args blip.CollectorFactoryArgs, frequency int) (blip.Collector, error) {
 	r.Lock()
 	defer r.Unlock()
 	f, ok := r.factory[domain]
@@ -83,7 +83,14 @@ func Make(domain string, args blip.CollectorFactoryArgs) (blip.Collector, error)
 		return nil, fmt.Errorf("invalid domain: %s (no factory registered)", domain)
 
 	}
-	return f.Make(domain, args)
+
+	c, err := f.Make(domain, args)
+
+	if err != nil {
+		return nil, err
+	}
+
+	return blip.NewCollectorFrequencyWrapper(c, frequency), nil
 }
 
 func PrintDomains() string {
@@ -97,7 +104,7 @@ func PrintDomains() string {
 
 	out := ""
 	for _, domain := range domains {
-		mc, _ := Make(domain, blip.CollectorFactoryArgs{Validate: true})
+		mc, _ := Make(domain, blip.CollectorFactoryArgs{Validate: true}, 0)
 		help := mc.Help()
 		out += fmt.Sprintf("%s\n\t%s\n\n",
 			help.Domain, help.Description,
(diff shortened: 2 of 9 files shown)


sink/datadog.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/DataDog/datadog-api-client-go/v2/api/datadog"
+	"github.com/DataDog/datadog-api-client-go/v2/api/datadogV2"
+
+	"github.com/cashapp/blip"
+	"github.com/cashapp/blip/sink/tr"
+	"github.com/cashapp/blip/status"
+)
+
+// SignalFx sends metrics to SignalFx.
+type DataDog struct {
```
daniel-nichter, line 20: `s/DataDog/Datadog/` as the latter is how they capitalize their name.

sink/datadog.go
```diff
@@ -0,0 +1,251 @@
...
+	monitorId string
+	dim       []string            // monitor.tags (dimensions)
```
daniel-nichter, line 22: If Datadog calls these tags, let's rename the var because "dim" is short for "dimension" which is what SFX calls them.

blip.go
```diff
@@ -65,6 +65,10 @@ type MetricValue struct {
 
 	// Meta is optional key-value pairs that annotate or describe the metric value.
 	Meta map[string]string
+
+	// The frequency that this metric is collected at. Some sinks need this value
+	// to properly process COUNTER type metrics.
+	Frequency int
```
daniel-nichter, line 71: There's a better, cleaner way to achieve this: add `Frequency` to `blip.Metrics` struct then set it at https://github.com/cashapp/blip/blob/main/monitor/engine.go#L273-L279

Then you can remove `collectorFrequencyWrapper` and everything related.

daniel-nichter (changes requested):

blip.go
Areson: Where we getting the `Frequency` value to set it there? We could pass it through as an arg to `engine.Collect` but the level info we end up with from [level_collector.go](https://github.com/cashapp/blip/blob/main/monitor/level_collector.go#L172-L177) is not necessarily the frequency that the collector is running at. For example, we have collectors that run every 30 seconds and others that run every 5 minutes. When the 5 minute collectors run the 30 second collectors also run, but the level that is pulled in `level_collector.go` is the index for the 5 minute collector, which is wrong for the 30 second collectors.

Areson: What about updating `atLevel` in the `Engine` [struct](https://github.com/cashapp/blip/blob/main/monitor/engine.go#L37-L58) to contain a wrapper type instead of just `Collector`? We could attach the frequency there and then use it at the suggested location without needing `collectorFrequencyWrapper`.

Areson: I take that back. I don't think that `Frequency` can live on `blip.Metrics` as that houses `MetricValue`s from multiple collectors that could potentially have different frequencies associated with them.

daniel-nichter: Hm, yes, metrics "level up" in overlapping frequencies, so a metric collected every 10s is collected every 60s, too.

Are you certain Datadog needs a frequency? I've seen/used many metrics systems, but I've never seen one require a frequency to calculate counters. (I'm kind of new to Datadog, though.) It doesn't make sense because it's possible and common to calculate counters (or rates from counters) without knowing the configured collection frequency.

Areson: I played around some more with this in the UI by removing the interval setting on data. The raw metrics for `COUNT` types still display, but any rate calculations fail to display at higher resolutions (appears to be sub-hour?). Putting the interval back fixes the issue. It looks like the interval dictates how DD displays things. If I set the interval to something more frequent than how often we collect the data points then the same issue occurs. I can set it more course (e.g. 60 seconds rather than 30) at which point the resolution of the rate calculations matches the interval, meaning it displays a data point every 60 seconds when zoomed in.

Areson: _Technically_ we can set this interval (and other metadata) in the UI without sending it over the UI. It means manually adjusting these values for each new metric we bring online. The interval data is shared across all of the metrics with the same name, so if we had a conflict with how the frequency was configured across blip instances that would result in the interval value ping-ponging between different values, which would be less than ideal. Part of me thinks that having the interval, units, etc. get set by the sink based on the configuration would be convenient but it's not a show stopper.

metrics/factory.go
```diff
@@ -75,14 +75,15 @@ func Exists(domain string) bool {
 // Make makes a metric collector for the domain using a previously registered factory.
 //
 // See types in the blip package for more details.
-func Make(domain string, args blip.CollectorFactoryArgs) (blip.Collector, error) {
+func Make(domain string, args blip.CollectorFactoryArgs, frequency int) (blip.Collector, error) {
```
daniel-nichter, line 78: We can get rid of `frequency` and all the new plumbing too, right?

Areson: Shoot, yes.

Areson: All cleaned up.

daniel-nichter (approved):

PR #69, opened 2022-10-03
Add monitor reloading

Areson:
Adding code to reload monitors via the API. Per comments in `loader.go` above `StartMonitors` it _seems_ like this functionality should exist. `loader.Load` supports detecting diffs in monitors but there is no hook from the API to run `Load` then `StartMonitors`.

Added a `Stop` function to `Engine` to facilitate cleaning up metrics when a monitor stops. The level collector has code that runs when a monitor stops but no cleanup on the engine, and the metrics it contains, is ever run. This leaves a heartbeat reader hanging for old monitors using the `repl.lag` metric.

The goal of adding this API endpoint is to facilitate the use of `plugins.LoadMonitors` where the plugin can dynamically update the monitors that need to run. Once the plugin updates the list of monitors it can invoke the reload by hitting the `/monitors/reload` endpoint.

### DIFF
--- a/monitor/engine.go
+++ b/monitor/engine.go
@@ -356,3 +356,21 @@ func (e *Engine) Collect(ctx context.Context, levelName string) (*blip.Metrics, 
 	atomic.AddUint64(&e.collectFail, 1)
 	return nil, fmt.Errorf("failed to collect %s/%s", e.plan.Name, levelName)
 }
+
+// Stop the engine and cleanup any metrics associated with it.
+func (e *Engine) Stop() {
+	blip.Debug("Stopping engine...")
+	e.planMux.Lock()
+	defer e.planMux.Unlock()
+
+	e.mcMux.Lock()
+	defer e.mcMux.Unlock()
+
+	// Clean up the monitors
+	for _, mc := range e.mcList {
+		if mc.cleanup != nil {
+			blip.Debug("%s cleanup", mc.c.Domain())
+			mc.cleanup()
+		}
+	}
+}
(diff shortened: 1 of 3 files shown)


monitor/engine.go
```diff
@@ -356,3 +356,21 @@ func (e *Engine) Collect(ctx context.Context, levelName string) (*blip.Metrics,
 	atomic.AddUint64(&e.collectFail, 1)
 	return nil, fmt.Errorf("failed to collect %s/%s", e.plan.Name, levelName)
 }
+
+// Stop the engine and cleanup any metrics associated with it.
+func (e *Engine) Stop() {
+	blip.Debug("Stopping engine...")
+	e.planMux.Lock()
+	defer e.planMux.Unlock()
```
daniel-nichter, line 364: There's a subtle race condition that we can ignore for now, but let's at least code comment it in this func so future engineers know that we know. If `Engine.Collect` had just been called, then monitor is stopped, it's possible this func locks `planMux` first and blocks the `Engine.Collect` goroutine. Then this func calls the cleanup func that could invalidate something in a collector. After this func returns, `Engline.Collect` can acquire the lock on `planMux` but the cleanup func has already been called, so the collector might crash. But that's ok: the panic will be caught and the monitor is being stopped anywhere. We can just note this potential but unlikely race condition.

Areson: Would something like having `LevelCollector` wait on the `sem` be sufficient to avoid this? Something like:
```go
case <-stopChan:
   // <other stop code>
   for i := 0; i < maxCollectors; i++ {
	<-sem
   }
   
   c.engine.Stop()
   return nil
```

PR #72, opened 2022-10-26
 Hotswappable Credentials

Areson:
Added support for full credentials (username, password, and/or TLS config) to DSN reload for driver. In some cases we have the contents of a CNF file updated with both new credentials and a new user name, and possibly updated certs. This supports that case while preserving the original functionality around a fixed username and rotated passwords.

This doest _not_ support turning TLS on/off dynamically. This approach assumes that once we choose to use TLS for a monitor that we will continue to do so, and we will only rotate the certs.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1017,6 +1017,7 @@ func (c ConfigTLS) Set() bool {
 	return !True(c.Disable) && c.MySQLMode != "DISABLED" && (c.CA != "" || c.Cert != "" || c.Key != "")
 }
 
+// Create tls.Config from the Blip TLS config settings. Returns
 func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
 	//  WARNING: ConfigTLS.Valid must be called first!
 	Debug("TLS for %s: %+v", c, server)
@@ -1038,6 +1039,7 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
 		if err != nil {
 			return nil, err
 		}
+
 		caCertPool := x509.NewCertPool()
 		caCertPool.AppendCertsFromPEM(caCert)
 		tlsConfig.RootCAs = caCertPool
@@ -1049,6 +1051,7 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
 		if err != nil {
 			return nil, err
 		}
+
 		tlsConfig.Certificates = []tls.Certificate{cert}
 		tlsConfig.BuildNameToCertificate()
 	}
--- a/dbconn/factory.go
+++ b/dbconn/factory.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io/fs"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/exec"
 	"regexp"
@@ -53,6 +54,11 @@ func (f factory) Make(cfg blip.ConfigMonitor) (*sql.DB, string, error) {
 	// ----------------------------------------------------------------------
 	// my.cnf
 
+	// Make a copy of the config before modifying it so that it can be used later,
+	// specifically for reloading TLS where we need to determine where the updates
+	// should come from
+	originalCfg := cfg
+
 	// Set values in cfg blip.ConfigMonitor from values in my.cnf. This does
 	// not overwrite any values in cfg already set. For exmaple, if username
 	// is specified in both, the default my.cnf username is ignored and the
@@ -82,29 +88,29 @@ func (f factory) Make(cfg blip.ConfigMonitor) (*sql.DB, string, error) {
 	// ----------------------------------------------------------------------
 	// Pasword reload func
 
-	// Blip presumes that passwords are rotated for security. So we create
-	// a callback that relaods the password based on its method: static, file,
+	// Blip presumes that credentials are rotated for security. So we create
+	// a callback that relaods the credentials based on its method: static, file,
 	// Amazon IAM auth token, etc. The special mysql-hotswap-dsn driver (below)
 	// calls this func when MySQL returns an authentication error.
-	passwordFunc, err := f.Password(cfg)
+	credentialFunc, err := f.Credentials(cfg)
 	if err != nil {
 		return nil, "", err
 	}
 
-	// Test the password reload func, i.e. get the current password, which
-	// might just be a static password in the Blip config file or another file,
+	// Test the credentials reload func, i.e. get the current credentials, which
+	// might just be a static credentials in the Blip config file or another file,
 	// but it could be something dynamic like an Amazon IAM auth token.
 	ctx, cancel := context.WithTimeout(context.Background(), 500*time.Millisecond)
 	defer cancel()
-	password, err := passwordFunc(ctx)
+	credentials, err := credentialFunc(ctx)
 	if err != nil {
 		return nil, "", err
 	}
 
 	// Credentials are username:password--part of the DSN created below
-	cred := cfg.Username
-	if password != "" {
-		cred += ":" + password
+	cred := credentials.Username
+	if credentials.Password != "" {
+		cred += ":" + credentials.Password
 	}
 
 	// ----------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


dbconn/factory.go
```diff
@@ -218,7 +229,19 @@ func (f factory) Password(cfg blip.ConfigMonitor) (PasswordFunc, error) {
 			return nil, err
 		}
 		token := aws.NewAuthToken(cfg.Username, cfg.Hostname, awscfg)
-		return token.Password, nil
+		return func(ctx context.Context) (*Credentials, error) {
+			passwd, err := token.Password(ctx)
+
```
daniel-nichter, line 234: Remove this empty line. The `if err != nil {` boilerplate is pretty much always next to the call that made `err`.

dbconn/reload_credentials.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+type Credentials struct {
+	Hostname string
+	Username string
+	Password string
+	TLS      *blip.ConfigTLS
+}
+
+type CredentialFunc func(context.Context) (*Credentials, error)
```
daniel-nichter, line 28: https://hackmysql.com/golang/go-antipatterns/#writable-read-only-value

`Credentials` are definitely not something we want code to accidentally/inadvertently modify because the same pointer is passed around.

dbconn/reload_credentials.go
```diff
@@ -0,0 +1,137 @@
...
+
+func init() {
+	dsndriver.SetHotswapFunc(Repo.ReloadDSN)
+}
+
+type Credentials struct {
+	Hostname string
+	Username string
+	Password string
+	TLS      *blip.ConfigTLS
```
daniel-nichter, line 25: https://hackmysql.com/golang/go-antipatterns/#writable-read-only-value

dbconn/reload_credentials.go
```diff
@@ -0,0 +1,137 @@
...
+	m       *sync.Map
+	tlsHash *sync.Map
+}
+
+var Repo = &repo{
+	m:       &sync.Map{},
+	tlsHash: &sync.Map{},
+}
+
+func (r *repo) Add(addr string, f CredentialFunc, tlsHash string) error {
```
daniel-nichter, line 40: Why is `tlsHash` not a field in `Credentials`? Seems like it's just another part of the credentials.

dbconn/reload_credentials.go
```diff
@@ -0,0 +1,137 @@
+// Copyright 2022 Block, Inc.
```
daniel-nichter, line 1: Why wasn't this file modified in place? The git diff is lost, and sucks because `ReloadDSN` is critical function.

daniel-nichter (changes requested):

config.go
```diff
@@ -1055,3 +1057,37 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
...
+	if c.CA != "" {
+		files = append(files, c.CA)
+	}
+
+	if c.Cert != "" && c.Key != "" {
+		files = append(files, c.Cert)
+		files = append(files, c.Key)
+	}
+
+	if len(files) > 0 {
```
daniel-nichter, line 1078: Go "reads down the left", so you want:
```
if len(files) == 0 {
    return "", nil
}

// Do the normal stuff
```

config.go
```diff
@@ -1055,3 +1057,37 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
 
 	return tlsConfig, nil
 }
+
+// Computes a hash of the cert files in the config
+func (c ConfigTLS) GetTLSHash() (string, error) {
```
daniel-nichter, line 1062: Also, why not compute the hash once, when the TLS file is loaded? Then store that has with the TLS config. (I've been away from this part of Blip code too long now, so I don't recall if this will work, but it seems like it should if the `ConfigTLS` struct is kept anywhere.)

dbconn/reload_credentials.go
daniel-nichter: Another not to recreate files: git history is lost. So if someone wants to see "who wrote this code?", I'll all point to you.

dbconn/reload_credentials.go
Areson: I could add the hash as a field on `Credentials` so it's computed when pulling the password info, if needed. For the `Add` method I'm leaning away from invoking `CredentialsFunc` to get the hash though, which would mean I'd leave the function signature as-is. Do you have a preference here?

config.go
Areson: I can compute the hash when loading the TLS config, but I don't have a place to put it. `blip.TLSConfig` is passed as a struct for the `LoadTLS` function so I can't update it. I can return an additional value from that function that contains the hash. Really that saves reading the files twice but I'm not sure if it makes things too much cleaner.

dbconn/reload_credentials.go
Areson: Alternatively, if we want the hash to be computed when loading the certs then this makes less sense as `CredentialsFunc` is just parsing the CNF file, not loading config based on the CNF file contents. To me feels like it makes more sense to have `CredentialsFunc` load the CNF file contents, which we then can check for changes in `ReloadDSN`. If we detect changes we load the TLS files. Granted we do have to parse the TLS files to generate the hash, but we could simply do that and load the TLS settings from the files at the same time and discard them if there are no changes.

config.go
Areson: I've updated the code to try and blend the suggestions here. The hash is computed while loading the `tls.Config` but returned as another return value. This cuts down on calls a bit. I'm not putting it into the `Credentials` struct at this point to avoid loading the certs as part of `CredentialsFunc` running, though this certainly could be changed.

config.go
```diff
@@ -1032,26 +1037,41 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
...
-		caCert, err := ioutil.ReadFile(c.CA)
+		f, err := os.Open(c.CA)
 		if err != nil {
-			return nil, err
+			return nil, "", err
 		}
+
+		data := io.TeeReader(f, hash)
+		caCert := new(bytes.Buffer)
+		caCert.ReadFrom(data)
```
daniel-nichter, line 1054: You sure these changes are needed? I think you can do:
```
hash.Write(caCert)

for _, bytes := range cert.Certificate {
	hash.Write(bytes)
}
```

config.go
```diff
@@ -1032,26 +1037,41 @@ func (c ConfigTLS) LoadTLS(server string) (*tls.Config, error) {
...
 		}
+
+		for _, bytes := range cert.Certificate {
+			hash.Write(bytes)
+		}
+
 		tlsConfig.Certificates = []tls.Certificate{cert}
 		tlsConfig.BuildNameToCertificate()
-	}
+	}
```
daniel-nichter, line 1074: Extra space probably means `go fmt` wasn't run.

config.go
Areson: Yup.

dbconn/factory.go
```diff
@@ -122,15 +123,58 @@ func (f factory) Make(cfg blip.ConfigMonitor) (*sql.DB, string, error) {
...
+				tls.SkipVerify = cfg.TLS.SkipVerify
+			}
+
+			tlsConfig, err := tls.LoadTLS(portSuffix.ReplaceAllString(cfg.Hostname, ""))
+
+			if err != nil {
+				// Don't interrupt the password reload if we have an error loading TLS.
+				// If there was a change we will get an error when trying to re-connect
+				// so log the error and continue
+				log.Printf(fmt.Sprintf("Error reloading TLS settings: %v", err))
```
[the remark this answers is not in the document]
Areson: Doh.

dbconn/factory.go
```diff
@@ -122,15 +123,58 @@ func (f factory) Make(cfg blip.ConfigMonitor) (*sql.DB, string, error) {
...
+			if err != nil {
+				return creds, err
+			}
+
+			// Reload from the original configuration by default
+			var tls blip.ConfigTLS = cfg.TLS
+
+			// If the credentials function returned back a config, use that instead
+			if creds.TLS.Set() {
+				tls = creds.TLS
```
daniel-nichter, line 155: This is too confusing. The confusion is `config.tls` (or `config.monitors.tls`) vs. `Credentials.TLS`: which one's the "real" TLS and why might one be set by not the other? (So I'd call this a case of DRY violation.) I think the latter only comes from a my.cnf? But at some point the code needs to DRY and make 1 TLS the authoritative one.

daniel-nichter (approved):

PR #73, opened 2022-10-26
Add support for Datadog metric resources

Areson:
Datadog supports associating infrastructure with metrics so that those relationships can be utilized in the UI. Added an association between the host and a metric based on if the "host" tag is present.

### DIFF
--- a/sink/datadog.go
+++ b/sink/datadog.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/DataDog/datadog-api-client-go/v2/api/datadog"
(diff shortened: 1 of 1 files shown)


sink/datadog.go
```diff
@@ -175,6 +176,24 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 				}
 			}
 
+			// If we have a "host" tag, we should include a resource definition for host
+			// so that metrics are properly associated with infrastructure in Datadog
+			var resources []datadogV2.MetricResource = nil
+
+			for _, tag := range s.tags {
+				if strings.Index(tag, "host:") == 0 {
```
daniel-nichter, line 184: `strings.HasPrefix`

daniel-nichter: Also, do this once in `NewDatadog` because tags don't change.

daniel-nichter (changes requested):

sink/datadog.go
Areson: Done.

daniel-nichter (approved):

PR #79, opened 2022-11-30
Fix query response time collectors

daniel-nichter:
Specify percentiles to collect in metrics, not options.
Instead of:
```yaml
query.response-time:
  options:
    percentiles: "95,99"
```
Now it's consistent with the design of Blip:
```yaml
query.response-time:
  metrics:
    - p95
    - p99
```

### DIFF
--- a/docs/v1.0/metrics/reporting.md
+++ b/docs/v1.0/metrics/reporting.md
@@ -27,7 +27,7 @@ Blip metric types are standard:
 
 * `COUNTER`
 * `GAUGE`
-* `BOOL`  (reserved for future use)
+* `BOOL`
 * `EVENT` (reserved for future use)
 
 Blip automatically uses the correct metric type for all metrics.
@@ -60,7 +60,7 @@ When the MySQL metric unit is variable, Blip uses the following units:
 |Data size|bytes
 
 For example, query response time ranges from nanoseconds to seconds (with microsecond precision) depending on the source.
-But regardless of the source, Blip reports `query.*.response_time` as _microseconds (μs)_.
+But regardless of the source, Blip reports query time as _microseconds (μs)_.
 
 {: .note}
 To convert units, use the [TransformMetrics plugin](../develop/integration-api#plugins) or write a [custom sink](../develop/sinks)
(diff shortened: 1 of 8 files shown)


docs/v1.0/metrics/reporting.md
```diff
@@ -89,42 +89,38 @@ type Metrics struct {
...
 ```
 
-All metrics belong to a single [domain](domains).
-Each domain collector can collect any number of metrics.
-Each metric has a type and value.
-[Groups](#groups) and [Meta](#meta) are detailed below.
-
 {: .note }
-Blip metrics are not stored or report in YAML; these examples are only for illustration.
+**Blip metrics are not stored or report in YAML. Examples are for illustration only.**
```
Areson, line 111:
```suggestion
**Blip metrics are not stored or reported in YAML. Examples are for illustration only.**
```

sqlutil/percentile.go
```diff
@@ -45,3 +47,26 @@ func FormatPercentile(f float64) string {
 	metaKey = "p" + metaKey
 	return metaKey
 }
+
+type P struct {
+	Name  string
+	Value float64
+}
+
+func PercentileMetrics(metrics []string) ([]P, error) {
```
wen-hui-sun, line 56: Would be better to have test case for it.

sqlutil/percentile.go
```diff
@@ -12,24 +12,26 @@ import (
...
 	var percentile float64
 	if f < 1 {
 		// percentile of the form 0.999 (P99.9)
 		percentile = f
-	} else if f >= 1 && f <= 100 {
+	} else if f >= 1 && f < 100 {
 		// percentile of the form 99.9 (P99.9)
 		percentile = f / 100.0
+	} else if f == 100.0 {
+		return 100, nil
```
wen-hui-sun, line 29: Shouldn't return 1 when f == 100.0?

wen-hui-sun: I think P1 should be 0.01 after parsing, and P100 should be 1 in this case. This result is used here: https://github.com/cashapp/blip/blob/main/metrics/percona/mysql_qrt_histogram.go#L53 and it is supposed to be in range of [0, 1.0]

daniel-nichter: Yep, you're right; I just saw that too. Fixing and testing.

sqlutil/percentile.go
daniel-nichter: Tests pushed.

wen-hui-sun (approved):

PR #80, opened 2022-12-05
Add dogStatsD sink

joycse06:
This is variant 1 where `dogstatsd` is implemented as it's own sink. This can be a bit confusing having two Sinks for Datadog (as `dogstatsD` is specialised to Datadog's statsD and won't work with other `statsD` enhancements).

I am exploring another option of extending the existing Datadog sink with `dogStatsD` in another PR.


daniel-nichter:
Hm, yeah. There's something to say for "different methods, different sinks", but also something to say for "same destination, same sink". In this case, user probably expects that the `datadog` sink goes to Datadog any way that DD supports. And code-wise I think that works because the options for the two are separate, so we won't clutter the single DD sink with too many options and capabilities.

joycse06:
Agreed. Extending the existing sink is the right thing to do here. I will close this off.

PR #81, opened 2022-12-06
Extend datadog sink with dogStatsD

joycse06:
This extends the `datadog` sink to use `dogStatsD` instead of api to send metrics to datadog as well. This is an alternative approach to https://github.com/cashapp/blip/pull/80

I think this one is better as it will only have one way of sending metrics to Datadog either via the API or DogStatsD.


sink/datadog.go
```diff
@@ -18,25 +20,27 @@ import (
 	"github.com/cashapp/blip/status"
 )
 
-// SignalFx sends metrics to SignalFx.
```
YinanLiang, line 21: 🙏

sink/datadog.go
```diff
@@ -92,29 +95,155 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+func (s *Datadog) sendDogStatsD(ctx context.Context, m *blip.Metrics) error {
+	status.Monitor(s.monitorId, s.Name(), "sending metrics")
+
+	// On return, set monitor status for this sink
+	n := 0
+	defer func() {
+		status.Monitor(s.monitorId, s.Name(), "last sent %d metrics at %d", n, time.Now())
+	}()
+
+	// Pre-alloc DogStatsD data points
```
daniel-nichter, line 163: None of this block is used. Looks like Dogstatd fires off every metric when one by one?

sink/datadog.go
```diff
@@ -92,29 +95,155 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+
+	// Convert each Blip metric value to an DogStatsD data point
+	for domain := range m.Values { // each domain
+		metrics := m.Values[domain]
+		var name string
+
+	METRICS:
+		for i := range metrics { // each metric in this domain
+
+			// Set full metric name: translator (if any) else Blip standard,
```
daniel-nichter, line 180: All this stuff in this loop is a big DRY violation with the other func that sends via API. So rather than duplicating code, let's add

```go
if s.dogstatsd {
   s.dogStatsDClient.Count(name, int64(metrics[i].Value), tags, 1)
   // ...
}
```

Right inline with the existing code.

Then update the docs to note that both can be used at same time, and dogstatsd is sent first, followed by API.

daniel-nichter (changes requested):

sink/datadog.go
```diff
@@ -92,29 +95,155 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
 			d.tr = tr
+
 		case "metric-prefix":
 			if v == "" {
 				return nil, fmt.Errorf("datadog sink metric-prefix is empty string; value required when option is specified")
 			}
 			d.prefix = v
 
+		case "use-dogstatsd":
+			d.useDogStatsD = blip.Bool(v)
```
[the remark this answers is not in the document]
joycse06: hmm, I thought about it but decided to make the intention to use `dogStatsD` explicit. I also think it's sensible to only use one of these options to send metrics to `datadog` not both at the same time (as that will create duplicate entries and mess things up) and I think blip shouldn't let the user do that. So I wanted to depend on something explicit to choose which path to take.

sink/datadog.go
joycse06: I was aware of the duplication, but I was trying to avoid branching, while there's opportunity fro DRYing up, there's also differences (the main one being one sends inline in the loop, another uses batching) in they way they work that will introduce branching. the `send` method is already pretty long, inlining the other path will make it even more long and complex. I will have a look and see what I can come up with.

sink/datadog.go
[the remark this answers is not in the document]
joycse06: I am still of the opinion, we shouldn't let the user use both `dogStatsD` and `api` at the same time as `dogStatsD` can only send to Datadog which means all the differences between these two paths have to be wrapped with `if {` switches. I can still find the intention of the user (if they want to use `api` or `dogstatsd`) from presence of `dogStatsD` or `api` options, but I would like to ideally put a validation that they haven't set up both.

sink/datadog.go
joycse06: Yeah you are right, I was also hoping there will be some kind of batching API but there's none. The go client does buffering by default and even does some basic client-side aggregation by default as well. I will fix it.

joycse06:
@daniel-nichter Attempted DRYing up the code. I think we shouldn't let user use both the API and DogStatsD at the same time.

I will update the doc once we have finalised the design.

sink/datadog.go
```diff
@@ -130,7 +160,11 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 	if n == 0 {
 		return fmt.Errorf("no Blip metrics were collected")
 	}
-	dp := make([]datadogV2.MetricSeries, n)
+	var dp []datadogV2.MetricSeries
+
+	if s.api() {
```
daniel-nichter, line 165: Don't use superfluous func like this (specifically here and generally speaking). It's just an unnecessary redirection for future eng. Instead,
```go
// Prealloc DD DP for API 
if !s.dogstatsd {
}
```
Plus, you do this below :-) `if s.dogStatsD {`

sink/datadog.go
```diff
@@ -18,25 +20,28 @@ import (
...
 	metricsApi *datadogV2.MetricsApi
 	apiKeyAuth string
 	appKeyAuth string
+	resources  []datadogV2.MetricResource
 
-	resources []datadogV2.MetricResource
+	// -- DogStatsD
+	dogStatsD       bool
+	dogStatsDClient *statsd.Client
+	dogStatsDHost   string
```
daniel-nichter, line 41: Generally you're correct to Caps a thing however the source does, and in this case I see DD does spell it like "DogStatsD". But since these are internal/private fields, the simpler `dogstatsd` is probably ok because I bet... yep, even in their own Go code example they lowercase it:
```go
dogstatsd_client, err := statsd.New("127.0.0.1:8125")
```

sink/datadog.go
```diff
@@ -92,23 +96,49 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
 
-	if d.appKeyAuth == "" {
-		return nil, fmt.Errorf("datadog sink required either app-key-auth or app-key-auth-file")
+	// if DogStatsD and api are both setup, return error as it will otherwise result in duplicate metrics
+	if d.dogStatsD && (d.apiKeyAuth != "" && d.appKeyAuth != "") {
+		return nil, fmt.Errorf("datadog sink requires either dogStatsD host or (api-key-auth and app-key-auth), not both at the same time")
+	}
+
+	if d.dogStatsD {
+		client, err := statsd.New(fmt.Sprintf("%s:%d", d.dogStatsDHost, DefaultDogStatsDPort))
```
daniel-nichter, line 124: You'll need to check and append port only if needed, like https://github.com/cashapp/blip/blob/main/aws/iam_auth.go#L23

sink/datadog.go
```diff
@@ -130,7 +160,11 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 	if n == 0 {
 		return fmt.Errorf("no Blip metrics were collected")
 	}
-	dp := make([]datadogV2.MetricSeries, n)
+	var dp []datadogV2.MetricSeries
+
+	if s.api() {
+		dp = make([]datadogV2.MetricSeries, n)
+	}
 	n = 0
```
daniel-nichter, line 168: I don't think `n` is used after here? If that's the case, then rm this line.

sink/datadog.go
joycse06: This is probably a bit too extreme, but superfluous well-named small functions often helps with reducing cognitive load of the future readers, as they won't have to do the expression evaluation in their head. For high level public functions, the general flow is often what you are interested in, not the details. I will inline it.

sink/datadog.go
joycse06: Will update 👍

daniel-nichter (approved):

joycse06:
Merging this, the doc for Datadog sink can probably be made better, but I have added the absolute minimal for now so I can move on.

PR #83, opened 2022-12-08
Adding `io.table` metrics

Areson:
- Added new metrics collection for table io metrics. These include row reads/write counts, plus wait time metrics. These are to replace the `percona` specific metrics with more generic MySQL metrics that are available in `5.7` and `8.0`.
- Minor editing to documentation

### DIFF
--- a/docs/v1.0/metrics/quick-ref.md
+++ b/docs/v1.0/metrics/quick-ref.md
@@ -40,6 +40,7 @@ The rest are reserved for future use.
 |ndb|MySQL NDB Cluster||
 |oracle|Oracle enhancements||
 |percona|Percona Server enhancements||
+|[`io.table`](domains#iotable)|Table IO metrics [`performance_schema.table_io_waits_summary_by_table`](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-table-wait-summary-tables.html#performance-schema-table-io-waits-summary-by-table-table)|v1.0.??|
 |[`percona.response-time`](domains#perconaresponse-time)|Percona Server 5.7 Response Time Distribution plugin|v1.0.0|
 |perconca.userstat|[Percona User Statistics](https://www.percona.com/doc/percona-server/8.0/diagnostics/user_stats.html)||
 |percona.userstat.index|Percona `userstat` index statistics (`INFORMATION_SCHEMA.INDEX_STATISTICS`)|
--- a/docs/v1.0/metrics/quick-ref.md
+++ b/docs/v1.0/metrics/quick-ref.md
@@ -40,7 +40,6 @@ The rest are reserved for future use.
 |ndb|MySQL NDB Cluster||
 |oracle|Oracle enhancements||
 |percona|Percona Server enhancements||
-|[`wait.io.table`](domains#iotable)|Table IO Wait metrics [`performance_schema.table_io_waits_summary_by_table`](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-table-wait-summary-tables.html#performance-schema-table-io-waits-summary-by-table-table)|v1.0.??|
 |[`percona.response-time`](domains#perconaresponse-time)|Percona Server 5.7 Response Time Distribution plugin|v1.0.0|
 |perconca.userstat|[Percona User Statistics](https://www.percona.com/doc/percona-server/8.0/diagnostics/user_stats.html)||
 |percona.userstat.index|Percona `userstat` index statistics (`INFORMATION_SCHEMA.INDEX_STATISTICS`)|
@@ -76,3 +75,4 @@ The rest are reserved for future use.
 |wait|Stage waits||
 |wait.current|Current waits||
 |wait.history|Historical waits||
+|[`wait.io.table`](domains#iotable)|Table IO Wait metrics [`performance_schema.table_io_waits_summary_by_table`](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-table-wait-summary-tables.html#performance-schema-table-io-waits-summary-by-table-table)|v1.0.0|
--- a/docs/v1.0/metrics/quick-ref.md
+++ b/docs/v1.0/metrics/quick-ref.md
@@ -40,7 +40,7 @@ The rest are reserved for future use.
 |ndb|MySQL NDB Cluster||
 |oracle|Oracle enhancements||
 |percona|Percona Server enhancements||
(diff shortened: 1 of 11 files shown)


Areson:
Opening this PR to start getting some initial feedback on adding a new metric, `io.table`. This metrics pulls data from `performance_schema.table_io_waits_summary_by_table` and is primarily meant to allows users to get table-level row metrics (reads/writes) but also includes wait statistics.

daniel-nichter:
Thank you for the PR and updating the docs, too!

Naming things is hard, so I'm thinking on this. My initial reaction is that "io" is too broad/generic (almost everything is I/O one way or another)... still thinking.

Areson:
Yeah. Given that the performance schema has a _lot_ of `io` type stuff, I figured this would need to change.

daniel-nichter:
@Areson Let's do these domains (where applicable here):
```
wait.io
        .database
        .table
        .index
        .file
wait.lock
        .mdl
        .table
        .row
```

So two new base domains `wait.io` and `wait.lock`.

Three interesting things about `table_io_waits_summary_by_table` and `table_io_waits_summary_by_index_usage`:

1. They're summary tables, so _not_ quite the same as the base from which they come: `events_waits_current`. For this latter, that's why we keep domain `wait.current`: this domain is for full, raw wait metrics.
2. These two summary tables are also a _subset_ of wait events: `wait/io/table/sql/handler`. So again, since they're not "all waits", that's why domain `wait.current` is reserved (and not used for these metrics).
3. The second is not actually index _usage_. Since these are wait events, they might not even occur (if there's no measurable wait). So we should this misleading term (i.e. index "usage"), especially because, iirc, Percona Server does expose true index usage counts.

Moreover, domains `wait.io` and `wait.table` mirror the subset of waits from which they come: `wait/io/table/sql/handler` and `wait/lock/table/sql/handler`.

Areson:
To confirm, this would mean this metric goes under `wait.io.table`?

daniel-nichter:
Correct, `wait.io.table`. (I see now your PR doesn't do/need ` wait.lock`.)

Areson:
I've updated this to the new domain name.

docs/v1.0/metrics/quick-ref.md
```diff
@@ -40,6 +40,7 @@ The rest are reserved for future use.
 |ndb|MySQL NDB Cluster||
 |oracle|Oracle enhancements||
 |percona|Percona Server enhancements||
+|[`wait.io.table`](domains#iotable)|Table IO Wait metrics [`performance_schema.table_io_waits_summary_by_table`](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-table-wait-summary-tables.html#performance-schema-table-io-waits-summary-by-table-table)|v1.0.??|
```
daniel-nichter, line 43: Move this to keep the list alphabetized.

metrics/wait.io.table/table.go
```diff
@@ -0,0 +1,236 @@
...
+		Description: "Table IO Waits",
+		Options: map[string]blip.CollectorHelpOption{
+			OPT_INCLUDE: {
+				Name: OPT_INCLUDE,
+				Desc: "Comma-separated list of database or table names to include (overrides option " + OPT_EXCLUDE + ")",
+			},
+			OPT_EXCLUDE: {
+				Name:    OPT_EXCLUDE,
+				Desc:    "Comma-separated list of database or table names to exclude (ignored if " + OPT_INCLUDE + " is set)",
+				Default: "mysql.*,information_schema.*,performance_schema.*,sys.*",
```
daniel-nichter, line 113: Let's put this string in a `const` because it's used below in
```go
dom.Options[OPT_EXCLUDE] = "mysql.*,information_schema.*,performance_schema.*,sys.*"
```
Without a single `const` to keep this DRY, some day some future eng will change one but not the other.

metrics/wait.io.table/table.go
```diff
@@ -0,0 +1,236 @@
...
+	OPT_EXCLUDE        = "exclude"
+	OPT_INCLUDE        = "include"
+	OPT_TRUNCATE_TABLE = "truncate-table"
+	OPT_ALL            = "all"
+
+	TRUNCATE_QUERY = "TRUNCATE TABLE performance_schema.table_io_waits_summary_by_table"
+)
+
+var (
+	metric_names = []string{
```
daniel-nichter, line 25: `CamelCase` is the norm in Go and Blip. We can also give this and `metic_map` below more intention-revealing names, maybe... `columnNames` (the list) and `columnExists` (the map since that's what it's used for: to see if it exists).

metrics/wait.io.table/query.go
```diff
@@ -0,0 +1,71 @@
...
+	if all, ok := set[OPT_ALL]; ok && strings.ToLower(all) == "yes" {
+		for _, name := range metric_names {
+			columns = append(columns, name)
+		}
+	} else {
+		// Default
+		for _, metric := range metrics {
+			metric = strings.ToLower(metric)
+
+			if _, ok := metric_map[metric]; ok {
```
daniel-nichter, line 35: Should it be an error if the column doesn't exist?

daniel-nichter (approved):

metrics/wait.io.table/query.go
Areson: I thought about it, but an error would mean that the metric would not collect at all and (possibly?) the plan would not load. Having a typo in a metric name is not show-stopping worthy in my name. At best we log it but don't throw an error. Really `TableIoQuery` doesn't throw _any_ errors so I should clean up the function definition.

metrics/wait.io.table/table.go
Areson: Do you want that to be a const in just this metric? We use this across multiple metrics.

daniel-nichter: Yeah, I've noticed this whole idea of include/exclude tables is popping up in several places. We can leave this here as-is for now, and later we can factor out the common stuff somewhere/somehow.

metrics/wait.io.table/query.go
daniel-nichter: Hm... yeah this is a part of Blip design that's not really nailed down: how are metric typos handled (since all/most metrics are explicitly listed)? We can leave this here as-is and deal with that question later.

daniel-nichter:
Kill that snake_caee, then I think it's ready to merge.

Areson:
> Kill that snake_caee, then I think it's ready to merge.

Where?

daniel-nichter:
Oh, nvm it's gone. I refreshed last time... but I think GitHub was still showing me old old. All good now. Let's merge it.

PR #88, opened 2023-01-03
Adding Partitioning to Datadog Sink

Areson:
Added options for compression and partitioning to the Datadog sink when the API method is used. When receiving an error related to the payload size being too large the sink will attempt to calculate a possible size that will satisfy the size requirements and re-send the metrics. In the event that it exhausts options and cannot send the metrics the attempt will fail.

Added the option to turn on compression for the API, which allows much larger payloads by default.

### DIFF
--- a/docs/v1.0/sinks/datadog.md
+++ b/docs/v1.0/sinks/datadog.md
@@ -32,4 +32,5 @@ sinks:
     metric-translator: ""
     metric-prefix: ""
     dogstatsd-host: ""
+    api-compress: "true"
 ```
--- a/sink/datadog.go
+++ b/sink/datadog.go
@@ -3,12 +3,18 @@
 package sink
 
 import (
+	"bytes"
+	"compress/zlib"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"net/http"
 	"regexp"
 	"strconv"
+	"strings"
+	"sync"
 	"time"
 
 	"github.com/DataDog/datadog-go/v5/statsd"
@@ -23,6 +29,10 @@ import (
 
 var portRe = regexp.MustCompile(`:\d+$`)
 
+const (
+	MAX_PAYLOAD_SIZE int = 512000
+)
+
 // Datadog sends metrics to Datadog.
 type Datadog struct {
 	monitorId string
@@ -35,6 +45,11 @@ type Datadog struct {
 	apiKeyAuth string
 	appKeyAuth string
 	resources  []datadogV2.MetricResource
+	compress   bool
+
+	maxMetricsPerRequest     int // Limit the number of metrics we send per request. Only used with the API
+	maxMetricsPerRequestLock sync.Mutex
+	maxPayloadSize           int
 
 	// -- DogStatsD
 	dogstatsd       bool
@@ -62,9 +77,12 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 	}
 
 	d := &Datadog{
-		monitorId: monitorId,
-		tags:      tagList,
-		resources: resources,
+		monitorId:            monitorId,
+		tags:                 tagList,
+		resources:            resources,
+		maxMetricsPerRequest: math.MaxInt32, // By default don't limit the number of metrics per request.
+		compress:             true,
+		maxPayloadSize:       MAX_PAYLOAD_SIZE,
 	}
 
 	for k, v := range opts {
@@ -104,6 +122,9 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 			}
 			d.prefix = v
 
+		case "api-compress":
+			d.compress = blip.Bool(v)
+
 		case "dogstatsd-host":
 			d.dogstatsdHost = v
 
@@ -141,6 +162,7 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 
 		c := datadog.NewConfiguration()
 		c.HTTPClient = httpClient
+		c.Compress = d.compress
 		metricsApi := datadogV2.NewMetricsApi(datadog.NewAPIClient(c))
 		d.metricsApi = metricsApi
 	}
@@ -170,6 +192,25 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 		dp = make([]datadogV2.MetricSeries, n)
 	}
 	n = 0
+
+	// Make a copy of maxMetricsPerRequest in case it gets updated by other threads
+	localMaxMetricsPerRequest := s.maxMetricsPerRequest
+	rangeStart := 0
+	var apiErrors []string
+
+	// Setup our context for API calls
+	ddCtx := context.WithValue(
+		ctx,
+		datadog.ContextAPIKeys,
+		map[string]datadog.APIKey{
+			"apiKeyAuth": {
+				Key: s.apiKeyAuth,
+			},
+			"appKeyAuth": {
+				Key: s.apiKeyAuth,
+			},
+		},
+	)
 
 	// Convert each Blip metric value to an SFX data point
 	for domain := range m.Values { // each domain
@@ -280,6 +321,14 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 			}
 
 			n++
+
+			// Check if we have reached the maximum number of metrics per request
+			if !s.dogstatsd && n%localMaxMetricsPerRequest == 0 {
+				if err := s.sendApi(ddCtx, dp[rangeStart:n]); err != nil {
+					apiErrors = append(apiErrors, err.Error())
+				}
+				rangeStart = n
+			}
 		} // metric
 	} // domain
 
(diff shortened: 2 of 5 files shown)


docs/v1.0/sinks/datadog.md
```diff
@@ -32,4 +32,11 @@ sinks:
...
     metric-prefix: ""
     dogstatsd-host: ""
+    api-use-compression: "false"
+    api-partition-size: ""
+    api-auto-partition: "true"
 ```
+
+## API Considerations
+
+When using the API directly the sink may run into issues when sending large payloads, as the Datadog API [restricts the payload to a maximum of 500Kib](https://docs.datadoghq.com/api/latest/metrics/). The DogStatsD option avoids this as it always sends metrics individually. By default the API option will attempt to automatically adjust the size of the payloads sent when it encounters an error related to a payload being too large. You can also explicitly set the number of metrics to send at once using `api-partition-size` and setting `api-auto-partition` to `false`. You can optionally turn on compression for the API, in which case Datadog will accept any payloads whose uncomressed size is less than 5Mib.
```
daniel-nichter, line 42: `500 KiB` -- bytes, not bits, I presume?

docs/v1.0/sinks/datadog.md
```diff
@@ -32,4 +32,5 @@ sinks:
     metric-translator: ""
     metric-prefix: ""
     dogstatsd-host: ""
-```
+    api-use-compression: "true"
```
daniel-nichter, line 35: `api-compression`  (or `api-compress` if you want to mimic the DD client). Blip doesn't use words like "enable", "use", etc. The only such on/off type word is the consistent prefix `disable-...`.

sink/datadog.go
```diff
@@ -104,6 +119,9 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 			}
 			d.prefix = v
 
+		case "api-use-compression":
+			d.compress = strings.ToLower(v) == "true"
```
daniel-nichter, line 123: `d.compress = bllip.Bool(v)`
https://pkg.go.dev/github.com/cashapp/blip#Bool

sink/datadog.go
```diff
@@ -141,13 +159,69 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+
+			if err = s.sendApi(ddCtx, dp[rangeStart:i]); err != nil {
+				pErr = fmt.Errorf("%s\n%s", pErr, err)
+			}
+		}
+
+		return pErr
+	}
+
+	if err != nil {
```
daniel-nichter, line 217: This needs to be handled before checking `r.StatusCode` because `err` will be a low-level error (e.g. network issue), so there's no `*http.Response`. Afaik, if err is not nil, then there must be an `*http.Response`.

sink/datadog.go
```diff
@@ -62,9 +75,11 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 	}
 
 	d := &Datadog{
-		monitorId: monitorId,
-		tags:      tagList,
-		resources: resources,
+		monitorId:     monitorId,
+		tags:          tagList,
+		resources:     resources,
+		partitionSize: math.MaxInt32,
```
daniel-nichter, line 81: Let's clarify "partition size" for future engineers who'll have to understand this code. The first confusing thing is partition (size of) _what_? We've got `MAX_PAYLOAD_SIZE` which is clear (max payload sizes are a common thing). So what's not clear is that `partitionSize` is actually something like "metrics per Send request", if I'm not mistaken? Then that plus starting with MaxInt32 makes it seems like we'll have to partition _a lot_ (when needed) because 2^32= ~4 _billion_, yet we only deal with metrics in the 10s or 1000s.

daniel-nichter (changes requested):

sink/datadog.go
```diff
@@ -293,33 +394,72 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
...
+	size, err := s.estimateSize(metric)
 	if err != nil {
-		blip.Debug("error sending data points to Datadog: %s", err)
-		blip.Debug("error sending data points to Datadog: Http Response - %s", r)
+		return err
 	}
 
-	return err
+	// Figure out how many metrics can fit inside the max payload, but pad it slightly to control for headers, etc.
+	estPartitionSize := (MAX_PAYLOAD_SIZE - 1000) / size
```
daniel-nichter, line 418: Here's another example of where partition size is confusing: comment says "how many _metrics_", but it's calculating a value in bytes (based off MAX_PAYLOAD_SIZE and current payload size). I don't see where code does something like "N metrics = B bytes, so new partition size is something fraction of N".

sink/datadog.go
```diff
@@ -293,33 +394,72 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
...
+
+	s.partitionLock.Lock()
+	defer s.partitionLock.Unlock()
+
+	if estPartitionSize >= s.partitionSize {
+		// The new partition size is greater than what we have now, so let's reduce it by a small factor to make sure we can send metrics properly
+		estPartitionSize = int(float32(s.partitionSize) * .9)
```
daniel-nichter, line 425: This comment/code will need some clarification for future engineers. What does "properly" mean? And why 0.9? Maybe something like:
```
// Reduce partition size by 10% as a heuristic/guess for finding the maximum partition size
// (metrics per send) that'll fit within the DD payload size limit
```

sink/datadog.go
Areson: I can update the name. My thought with starting with MaxInt32 was that we begin with _no_ partitioning of the metrics. We only start to do that if we happen to run into the case where we get a error response back from Datadog indicating the payload is too large. With compression on that is much less likely to happen in general. MaxInt32 is just a nice default to set this to instead of just arbitrarily picking a value I thought was big enough.

sink/datadog.go
```diff
@@ -141,13 +160,70 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+				if err = s.sendApi(ddCtx, dp[rangeStart:i]); err != nil {
+					pErr = fmt.Errorf("%s\n%s", pErr, err)
+				}
+			}
+
+			return pErr
+		}
+
+		blip.Debug("error sending data points to Datadog: %s", err)
+		blip.Debug("error sending data points to Datadog: Http Response - %s", r)
```
daniel-nichter, line 221: Combine these.

sink/datadog.go
```diff
@@ -141,13 +160,70 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+	}
+
+	return err
```
daniel-nichter, line 224: `return nil` because the `if` above already checked that it's not nil, so an explicit `return nil` is idiomatic to show that func has reached a successful end/return.

daniel-nichter: By contrast, this kind of `return err` is used when a func returns the last return val of its last call without checking that return value.

sink/datadog.go
daniel-nichter: Oh actually, you need a `return nil` here to handle other error so you can `return nil` below. I.e. as the code is written now, it's mixing boilerplates, which is confusing. I.e. the bloilerplat,
```go
err := f()
if err = nil {
    return err
}
```
isn't happening.

sink/datadog.go
```diff
@@ -141,13 +160,70 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
...
+		Series: dp,
+	}
+	_, r, err := s.metricsApi.SubmitMetrics(ddCtx, payload, *datadogV2.NewSubmitMetricsOptionalParameters())
+	// Check to see if the request was too large
+	if err != nil {
+		if r.StatusCode == http.StatusRequestEntityTooLarge {
+			oldMaxMetricsPerRequest := s.maxMetricsPerRequest
+
+			// Sample one of the metrics and try caculating a new maxMetricsPerRequest
+			if pErr := s.updateMaxMetricsPerRequest(dp[0], len(dp)); pErr != nil {
```
daniel-nichter, line 184: Don't use `pErr` here in a block scope but then again later in another (outer) block scope. They have same special name but are different, which contradicts the normal expectation that any specifically-named var is the same var wherever the specific name occurs. (This is not true for nonspecifically named vars like the ubiquitous `err` in Go, and other classics like `i` in loops, etc.)

test/mock/tr.go
```diff
@@ -0,0 +1,13 @@
+package mock
+
+type Tr struct {
+	TranslateF func(domain, metric string) string
```
daniel-nichter, line 4: Strike the trailing `F` so mock field name = func name. (Other mocks in Blip should follow this convention.)

daniel-nichter (approved):

daniel-nichter:
Squash commits please. GitHub does this thing where it generates release notes from git commits, which is super handy when doing releases. So we'll try to squash commits in the project to make cutting releases a little easier.

test/mock/tr.go
Areson: It looks like the other mocks favor using `Func` rather than just `F`. Match that then?

daniel-nichter: Correction (as you noted on Slack): "Func" suffix.

PR #89, opened 2023-01-03
Add OPT_REPORT_NOT_A_REPLICA option to repl.running

joycse06:
Certain metrics platform doesn't have the ability to filter out data points from raw MTS, so adding this option to drop the `repl.running` metric when the host is not a replica.

### DIFF
--- a/metrics/repl/repl.go
+++ b/metrics/repl/repl.go
@@ -20,6 +20,8 @@ const (
 	NOT_A_REPLICA = -1
 
 	ERR_NO_ACCESS = "access-denied"
+
+	OPT_REPORT_NOT_A_REPLICA = "report-not-a-replica"
 )
 
 type replMetrics struct {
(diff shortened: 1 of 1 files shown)


metrics/repl/repl.go
```diff
@@ -54,7 +58,17 @@ func (c *Repl) Help() blip.CollectorHelp {
 	return blip.CollectorHelp{
 		Domain:      DOMAIN,
 		Description: "Replication status",
-		Options:     map[string]blip.CollectorHelpOption{},
+		Options: map[string]blip.CollectorHelpOption{
+			OPT_REPORT_NOT_A_REPLICA: {
+				Name:    OPT_REPORT_NOT_A_REPLICA,
+				Desc:    "Report not a replica as -1",
+				Default: "yes",
```
joycse06, line 65: Setting it to `yes` to preserve current behavior.

daniel-nichter (approved):

PR #93, opened 2023-02-01
Moving Package Variable to Monitor

Areson:
Moved the package-level variables `statusQuery` and `newTerms` to the monitor in case multiple monitors within the same domain are created but are running on different MySQL versions.

Added a check around `replStatus` when getting the "source" entry. In cases where something is not a replica the map will be nil and should not be accessed.

### DIFF
--- a/metrics/repl/repl.go
+++ b/metrics/repl/repl.go
@@ -34,6 +34,8 @@ type Repl struct {
 	errPolicy       map[string]*errors.Policy
 	stop            bool
 	dropNotAReplica map[string]bool
+	statusQuery     string
+	newTerms        bool
 }
 
 var _ blip.Collector = &Repl{}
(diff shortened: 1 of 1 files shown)


metrics/repl/repl.go
```diff
@@ -86,9 +90,6 @@ func (c *Repl) Help() blip.CollectorHelp {
 	}
 }
 
-var statusQuery = "SHOW SLAVE STATUS" // SHOW REPLICA STATUS as of 8.022
```
daniel-nichter, line 89: Please put that comment back, i.e. (few lines up in the diff):
```go
		statusQuery:     "SHOW SLAVE STATUS",  // SHOW REPLICA STATUS as of 8.022
```

When working with MySQL, you'll find that such "as of <version>" comments are common and helpful, especially when your career with MySQL begins to span decades. Lots and lots of changes, and who can remember them all?

daniel-nichter (approved):

PR #103, opened 2023-04-21
Add error handling for errors in Datadog API response

joycse06:
Datadog can return a `202 Accepted` response code but errors in response payload. This PR adds a check for that.

From what it looks like, the endpoint probably also partially accept MetricData which are valid and return errors for invalid ones.

With this, we will at least see in logs what's going on and will be able to fix it quickly. Without this, it was failing silently.

### DIFF
--- a/sink/datadog.go
+++ b/sink/datadog.go
@@ -373,16 +373,17 @@ func (s *Datadog) sendApi(ddCtx context.Context, dp []datadogV2.MetricSeries) er
 			optParams.ContentEncoding = datadogV2.METRICCONTENTENCODING_GZIP.Ptr()
 		}
 
-		if _, r, err := s.metricsApi.SubmitMetrics(ddCtx, *datadogV2.NewMetricPayload(dp[rangeStart:rangeEnd]), optParams); err != nil {
+		apiResponse, r, err := s.metricsApi.SubmitMetrics(ddCtx, *datadogV2.NewMetricPayload(dp[rangeStart:rangeEnd]), optParams)
+		if err != nil {
 			if r != nil && r.StatusCode == http.StatusRequestEntityTooLarge {
 				// Is the number of metrics sent already the smallest possible?
 				if localMaxMetricsPerRequest == 1 {
-					return fmt.Errorf("Unable to send metrics: %v", err)
+					return fmt.Errorf("unable to send metrics: %v", err)
 				}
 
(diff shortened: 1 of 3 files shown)


daniel-nichter (approved):

Areson (commented):
What was the failure mode for this? From what it looks like it appear that it _didn't_ throw an error at all since `err == nil`. If some of the metrics are being accepted returning an error here for `202` responses may cause some metric values not to be submitted if there were a large number of points being submitted. Are we fine with stopping the submission of any remaining metrics when we get a partial success?

sink/datadog.go
```diff
@@ -393,6 +394,11 @@ func (s *Datadog) sendApi(ddCtx context.Context, dp []datadogV2.MetricSeries) er
 			return err
 		}
 
+		if len(apiResponse.Errors) > 0 {
+			// datadog can return a 202 Accepted response code but errors in response payload
+			return fmt.Errorf("error response from Datadog: %s", strings.Join(apiResponse.Errors, ","))
```
daniel-nichter, line 399: Re @Areson comment:  if this is _partial_ success, then we should not abort completely.

joycse06:
The series that was being sent all had the same issue, so datadog returned error for the whole batch, but I think it can have partial success. I will verify.

But I agree, if it returns partial success, we should log the error and continue. Though I prefer to fail fast, but in this case as `blip` submits a lot of metrics we shouldn't stop submission for a few failures.

It might be a good idea to send a meta metric for partial errors in `sinks` so we can build alert on them. I don't want to depend on a human being to check for partial errors in logs all the time.

joycse06:
> What was the failure mode for this?

The metrics were not being sent to Datadog and there were no errors in logs. So it was failing silently.

daniel-nichter:
> It might be a good idea to send a meta metric for partial errors in `sinks` so we can build alert on them. I don't want to depend on a human being to check for partial errors in logs all the time.

I thought about meta metrics, but it's one of those areas where probably nothing we do will be right for a general audience. So instead, partial failure should emit a Blip event because (my thinking was) those can be picked up by a custom sink and sent wherever, to do whatever.

joycse06:
Verified, this can be partial success. Updated the code to log the errors and continue.

joycse06:
Raised an event as well.

Areson (approved):

PR #104, opened 2023-04-26
Truncation Timeouts

Areson:
Added timeouts on truncation operations for `query.response-time` and `wait.io.table`. In MySQL [locking an instance for backup](https://dev.mysql.com/doc/refman/8.0/en/lock-instance-for-backup.html) will cause `TRUNCATE` to block. This will cause multiple instances of the command to stack up, driving up the thread count on the server. A default timeout of `2s` has been set and is configurable by the user. An error, `truncate-failed` has also been setup to allow users to specify an error handling strategy in the event that `TRUNCATE` fails.

### DIFF
--- a/docs/v1.0/metrics/domains.md
+++ b/docs/v1.0/metrics/domains.md
@@ -126,7 +126,7 @@ _MySQL Query Response Time_
 |Derived metrics|&bull; `pN` (gauge)<br>|
 |Meta|&bull; `pN=pA`: where `pN` is collected percentile and `pA` is actual percentile|
 |Options|&bull; `real-percentiles`<br>&bull; `truncate-table`<br>&bull; `truncate-timeout`|
-|Error policy|&bull; `table-not-exist`<br>&bull; `truncate-failed`|
+|Error policy|&bull; `table-not-exist`<br>&bull; `truncate-timeout`|
 
 The `query.response-time` domain collect query response time percentiles.
 By default, it reports the P999 (99.9th percentile) response time in microseconds.
@@ -164,7 +164,7 @@ This resets percentile values so that each collection represents the global quer
 However, truncating the table interferes with other tools reading (or truncating) the table.
 
 * `truncate-timeout`<br>
-Default: 2s<br>
+Default: 250ms<br>
 The amount of time to wait while attempting to truncate [performance_schema.events_statements_histogram_global](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-statement-histogram-summary-tables.html).
 
 #### Error Policy
@@ -173,7 +173,7 @@ The amount of time to wait while attempting to truncate [performance_schema.even
 * `table-not-exist`<br>
 MySQL error 1146: Table 'performance_schema.events_statements_histogram_global' doesn't exist
 
-* `truncate-failed`<br>
+* `truncate-timeout`<br>
 Truncation failures on table 'performance_schema.events_statements_histogram_global'
 
 <!-------------------------------------------------------------------------->
@@ -470,7 +470,7 @@ _Table I/O Wait Metrics_
 |MySQL config|yes|
 |Sources|`performance_schema.table_io_waits_summary_by_table`|
 |Options|&bull; `exclude`<br>&bull; `include`<br>&bull; `truncate`<br>&bull; `truncate-timeout`<br>&bull; `all`|
-|Error policy|&bull; `truncate-failed`|
+|Error policy|&bull; `truncate-timeout`|
 |Group keys|`db`, `tbl`|
 
 Summarized table I/O wait metrics from `performance_schema.table_io_waits_summary_by_table`.
@@ -491,7 +491,7 @@ Default: `yes`<br>
 If the source table should be truncated to reset data after each retrieval.
 
 * `truncate-timeout`<br>
-Default: 2s<br>
+Default: 250ms<br>
 The amount of time to wait while attempting to truncate [performance_schema.events_statements_histogram_global](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-statement-histogram-summary-tables.html).
 
 * `all`<br>
@@ -502,5 +502,5 @@ If `no` (the default), only the explicitly listed `performance_schema.table_io_w
 #### Error Policy
 {: .no_toc }
 
-* `truncate-failed`<br>
+* `truncate-timeout`<br>
 Truncation failures on table 'performance_schema.events_statements_histogram_global'
--- a/docs/v1.0/metrics/domains.md
+++ b/docs/v1.0/metrics/domains.md
@@ -125,8 +125,8 @@ _MySQL Query Response Time_
 |Sources|MySQL 8.0 [p_s.events_statements_histogram_global](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-statement-histogram-summary-tables.html)|
 |Derived metrics|&bull; `pN` (gauge)<br>|
 |Meta|&bull; `pN=pA`: where `pN` is collected percentile and `pA` is actual percentile|
-|Options|&bull; `real-percentiles`<br>&bull; `truncate-table`|
-|Error policy|&bull; `table-not-exist`|
+|Options|&bull; `real-percentiles`<br>&bull; `truncate-table`<br>&bull; `truncate-timeout`|
+|Error policy|&bull; `table-not-exist`<br>&bull; `truncate-failed`|
 
 The `query.response-time` domain collect query response time percentiles.
 By default, it reports the P999 (99.9th percentile) response time in microseconds.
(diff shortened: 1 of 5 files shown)


docs/v1.0/metrics/domains.md
```diff
@@ -158,17 +158,24 @@ Therefore, the P99 might actually be P98.9 or P99.2.
...
+Default: 2s<br>
+The amount of time to wait while attempting to truncate [performance_schema.events_statements_histogram_global](https://dev.mysql.com/doc/refman/8.0/en/performance-schema-statement-histogram-summary-tables.html).
+
 #### Error Policy
 {: .no_toc }
 
 * `table-not-exist`<br>
 MySQL error 1146: Table 'performance_schema.events_statements_histogram_global' doesn't exist
 
+* `truncate-failed`<br>
```
daniel-nichter, line 176: s//truncate-timeout/ so it matches the option, creating a nice symmetry: here's the timeout option and here's what to do if that timeout errors.

docs/v1.0/metrics/domains.md
```diff
@@ -482,7 +490,17 @@ A comma-separated list of database or table names to exclude (ignored if `includ
 Default: `yes`<br>
 If the source table should be truncated to reset data after each retrieval.
 
+* `truncate-timeout`<br>
+Default: 2s<br>
```
daniel-nichter, line 494: Let's be more conservative: 500ms. Truncate should be nearly instantaneous. Even 500ms is really long for this query. Other issue is that the whole level will wait for this, and 2s is a long wait (in this context).

daniel-nichter: Maybe even 250ms. I'd measure how long it normally takes then do 2x or 3x that, to use something quasi-objective.

metrics/query.response-time/response_time.go
```diff
@@ -179,15 +209,60 @@ func (c *ResponseTime) Collect(ctx context.Context, levelName string) ([]blip.Me
...
 		if err != nil {
-			return nil, err
+			return c.collectTruncateError(err, levelName, metrics)
+		} else if c.atLevel[levelName].hadTruncateError {
+			// Upon success we can start emitting metrics normally again but we need to
+			// wait for the next collect for the truncate to have taken effect.
+			// Emit whatever metrics would be emitted based on the error policy.
+			metrics, err = c.collectTruncateError(fmt.Errorf("Truncate successful, waiting for next metric collection."), levelName, metrics)
 		}
+		c.atLevel[levelName].hadTruncateError = false
```
daniel-nichter, line 223: I know what this stuff is doing, but it's a little hard to follow. It's something like:
1. error
2. success (don't report)
3. success (ok now start reporting)
I'd see if there's a way to make it more clear, either in code or with comments.

docs/v1.0/metrics/domains.md
Areson: Can do. To be clear, I'm triggering this on _any_ failure with the truncate, not just timeouts. A timeout is those most likely event, but really we want to avoid emitting metrics if we have truncation failures in general.

metrics/wait.io.table/table.go
```diff
@@ -223,11 +267,57 @@ func (t *Table) Collect(ctx context.Context, levelName string) ([]blip.MetricVal
...
+			metrics, err = t.collectTruncateError(fmt.Errorf("Truncate successful, waiting for next metric collection."), o, metrics)
 		}
+
+		o.hadTruncateError = false
 	}
 
 	return metrics, err
 }
+
+func (t *Table) collectTruncateError(err error, options *tableOptions, collectedMetrics []blip.MetricValue) ([]blip.MetricValue, error) {
```
daniel-nichter, line 288: Is it possible this func and the one for the other collector can be factored out into common code somewhere? No big deal for now; just an idea in case future collectors `TRUNCATE` and need this same error policy logic.

Areson: Yes. They are pretty much the same and I can make a common version. Is there a place that makes sense to put that?

docs/v1.0/metrics/domains.md
Areson: 60 - 80ms seems common. 250 seems like a reasonable default. We just need it to too many instances of the query stacking up, so as long as it is less than the sampling frequency it should be fine.

Areson:
I took the route of creating a struct to make the handling of the truncation logic a bit more self-contained and easy to use. I don’t love the need of passing a pointer to the stop variable in, but in order to let it modify that value when stop is the policy it is needed. The alternative was to implement an interface that collectors could use on their config structs that implemented a function like Stop but I thought that might be a bit heavier than we needed. I can pivot to using that instead.

errors/truncate_error.go
```diff
@@ -0,0 +1,71 @@
...
+		return collectedMetrics, nil
+	}
+	// We may need to change the resulting metric values depending on if we had an error with truncation or not.
+	// If this is the first instance of a truncation error then the metrics are fine as-is. Once truncation fails
+	// subsequent metrics will have accumulated in the `performance_schema` table over a different interval than
+	// prior samples. In those cases we want to apply our error policy for metrics to the collected values.
+	// We will continue to apply the error processing to metrics until truncation succeeds and the interval of the
+	// collected values returns to the expected value.
+	processMetrics := p.hadTruncateError
+	p.hadTruncateError = (err != nil)
```
daniel-nichter, line 37: Move this block closer to where it's used. E.g. processMetrics is defined here but then there are 2 more blocks until it's used. It make sense now (with code comment), but it'll achieve maximum "max sense" when it's not "interrupted" by other code blocks.

errors/truncate_error.go
```diff
@@ -0,0 +1,71 @@
+package errors
+
+import (
+	"github.com/cashapp/blip"
+)
+
+type TruncateErrorPolicy struct {
+	Policy
```
daniel-nichter, line 8: Don't embed; avoid embedding in general because it hides which method belongs to which struct, which is a bit of magic, and Go doesn't like magic. E.g. it didn't occur to me at first that `p.Metric` comes from `Policy`, not `TruncateErrorPolicy`--and I wrote `Policy` (but, granted, that was awhile ago).

errors/truncate_error.go
```diff
@@ -0,0 +1,71 @@
...
+	hadTruncateError bool
+}
+
+func NewTruncateErrorPolicy(s string) *TruncateErrorPolicy {
+	p := NewPolicy(s)
+	return &TruncateErrorPolicy{
+		Policy: *p,
```
daniel-nichter, line 15: No sure this is right because `NewPolicy(s string) *Policy`, so this would be `**p`, I think?

daniel-nichter: Or, maybe it's somehow needed/due to embedding?

metrics/query.response-time/response_time.go
```diff
@@ -179,10 +209,11 @@ func (c *ResponseTime) Collect(ctx context.Context, levelName string) ([]blip.Me
...
 
 	if c.atLevel[levelName].truncate {
-		_, err := c.db.Exec(TRUNCATE_QUERY)
-		if err != nil {
-			return nil, err
-		}
+		trCtx, cancelFn := context.WithTimeout(ctx, c.atLevel[levelName].truncateTimeout)
+		defer cancelFn()
+		_, err := c.db.ExecContext(trCtx, TRUNCATE_QUERY)
+		// Process any errors (or lack thereof) with the TruncateErrorPolicy.
```
daniel-nichter, line 215: Comment says what but not really why, which will be important for future engineers. You don't have to explain again, but something like:
```go
// Always process errors (or lack thereof) with the TruncateErrorPolicy because there's special handling
// that needs to be applied on error and not error. See comments in TruncateErrorPolicy for details.
```

That way some future engineer will really have to be out of their mind to read that yet ignore it and undo this logic because they (incorrectly) think that this is standard `err` handling, so it can be ignored if `err==nil`.

errors/truncate_error.go
Areson: I'm using it as a value rather than a reference, but the method returns a pointer, so I have to dereference it to get the value to assign it to the field. It passes both the unit tests and integration testing so I think we are good.

daniel-nichter: Ah right, I see what's going on. Idk why but I was thinking `**p` as pointer to pointer instead of Go deref.

daniel-nichter (approved):

metrics/wait.io.table/table.go
```diff
@@ -165,6 +188,34 @@ LEVEL:
...
+			// greater than our truncate timeout.
+			lockWaitTimeout := math.Ceil(o.truncateTimeout.Seconds())
+			if lockWaitTimeout < 1.0 {
+				lockWaitTimeout = 1
+			}
+
+			o.lockWaitQuery = fmt.Sprintf(LOCKWAIT_QUERY, int64(lockWaitTimeout))
+		}
+
+		if o.truncate {
```
daniel-nichter, line 214: Redundant with the same `if o.truncate` above.

daniel-nichter (approved):

PR #106, opened 2023-06-05
DD sink should send deltas for counters

joycse06:
Details on: https://docs.google.com/document/d/11ORtaMFBD_jxKrwd6Dj__nFt8pTrU5lyw60u_pfWAJo/edit#heading=h.apxh14sfucxy

tldr; Update the Datadog sink to send `deltas` for counter metrics rather than sending a current value of the counter.

This was working in SFX because SFX has a `cumulative counter` type, whereas Datadog expects `deltas`.


YinanLiang (approved):

Areson (changes requested):

sink/datadog.go
```diff
@@ -277,8 +279,23 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 			// Convert Blip metric type to Datadog metric type
 			switch metrics[i].Type {
 			case blip.COUNTER:
+				// calculate delta
+				val, ok := s.counters[name]
```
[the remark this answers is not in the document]
joycse06: Good point @Areson .

Both `size.database` and `size.table` are `guage` metrics so they won't be affected by this.

`query.response-time` is a `guage` as well.

However, `wait.io.table` is a counter. It's also a common use case we should handle.

I was not taking `tag` aspect of the metrics into consideration at all, my bad. With current implementation, it would alter/override values for different tags.

I can see we are building a `tags` array, I can probably just append the contents of that into the name to get a unique ID for the metric/tags combination? I will give it a go.

sink/datadog.go
```diff
@@ -487,6 +499,25 @@ func (s *Datadog) estimateSize(metrics []datadogV2.MetricSeries) (int, error) {
...
 
+func (s *Datadog) uniqIdentifierForMetric(name string, tags []string) string {
+	var key string
+	key += name
+	for _, t := range tags {
+		v := s.intForTag(t)
+		key += strconv.Itoa(int(v))
+	}
+	return key
+}
```
Areson, line 510: Are we concerned about tag values possibly being inconsistently ordered? We may want to consider sorting the tags prior to generating the key to ensure a consistent order.

sink/datadog.go
```diff
@@ -279,23 +285,29 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
...
-				}
-				// previous value present, calculate delta
-				delta := metrics[i].Value - val
-				s.counters[name] = metrics[i].Value
-				if delta < 0 {
-					blip.Debug("error calculating delta for the counter: %s, got negative delta (can happen due to restart), skipping this run (next run will work).", name)
-					continue METRICS
+				metricVal := metrics[i].Value
+
+				if s.sendCounterDelta && metrics[i].ValueType != blip.DELTA {
```
Areson, line 290: I propose we change this to 
```suggestion
				if s.sendCounterDelta && metrics[i].ValueType == blip.CUMULATIVE {
```

I'd lean in favor of leaving `UKNOWN` value unchanged.

sink/datadog.go
joycse06: I had that suspicion as well, although the code flow kind of indicate it should be consistently ordered but it's probably not worth the risk. It will add a bit more overhead on for every metric on every send. Should be alright. I will add it.

sink/datadog.go
joycse06: Will update, I was trying to avoid adding `CUMULATIVE` to all other call sites. But you are right, we shouldn't modify `UNKNOWN` values.

sink/datadog.go
```diff
@@ -487,6 +499,25 @@ func (s *Datadog) estimateSize(metrics []datadogV2.MetricSeries) (int, error) {
...
+
+func (s *Datadog) intForTag(tag string) uint16 {
+	v, ok := s.tagToInt[tag]
+
+	if !ok {
+		s.tagToInt[tag] = uint16(len(s.tagToInt))
+	}
+	return v
+}
+
```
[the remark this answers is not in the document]
joycse06: Thats better, I will update. For a lot of distinct tags the value can indeed get high. 👍

joycse06:
@Areson Applied your feedback, PTAL 🙏

Areson:
This generally looks good to me. I'm going to see if I can do some integration checking on my end.

Areson (approved):
Finished integration testing. The changes look good. We should add documentation about the change with notes about tagging with a blip version to track if desired.

joycse06:
Thanks for testing it out @Areson . I will work on documenting it today.

prom/tr/generic.go
```diff
@@ -36,9 +36,13 @@ func (tr Generic) Translate(values []blip.MetricValue, ch chan<- prometheus.Metr
 		var promType prometheus.ValueType
 		var help string
 		switch values[i].Type {
-		case blip.COUNTER:
+		case blip.CUMULATIVE_COUNTER:
 			promType = prometheus.CounterValue
 			help = "Generic counter metric."
+		case blip.DELTA_COUNTER:
```
joycse06, line 42: We could do the reverse we are doing in Datadog sink (maintaining a cumulative counter). I am leaving it for a future change?

prom/tr/innodb_metrics.go
```diff
@@ -61,9 +61,13 @@ func (tr InnoDBMetrics) Translate(values []blip.MetricValue, ch chan<- prom.Metr
 		var promType prom.ValueType
 		var help string
 		switch values[i].Type {
-		case blip.COUNTER:
+		case blip.CUMULATIVE_COUNTER:
 			promType = prom.CounterValue
 			help = "Generic counter metric from SHOW GLOBAL STATUS."
+		case blip.DELTA_COUNTER:
```
joycse06, line 67: Same here:

> We could do the reverse we are doing in Datadog sink (maintaining a cumulative counter). I am leaving it for a future change?

prom/tr/status.go
```diff
@@ -30,9 +30,13 @@ func (tr StatusGlobal) Translate(values []blip.MetricValue, ch chan<- prom.Metri
 		var promType prom.ValueType
 		var help string
 		switch values[i].Type {
-		case blip.COUNTER:
+		case blip.CUMULATIVE_COUNTER:
 			promType = prom.CounterValue
 			help = "Generic counter metric from SHOW GLOBAL STATUS."
+		case blip.DELTA_COUNTER:
```
joycse06, line 36: Same here:
> We could do the reverse we are doing in Datadog sink (maintaining a cumulative counter). I am leaving it for a future change?

sink/chronosphere.go
```diff
@@ -170,14 +171,18 @@ func (s *Chronosphere) Send(ctx context.Context, m *blip.Metrics) (lerr error) {
...
-			default: // COUNTER
+			case blip.CUMULATIVE_COUNTER:
 				fam[n].Metrics[0].MetricPoints[0].Value = &om.MetricPoint_CounterValue{
 					CounterValue: &om.CounterValue{
 						Total: &om.CounterValue_DoubleValue{
 							DoubleValue: m.Value, // VALUE (counter)
 						},
 					},
 				}
+			default:
```
joycse06, line 182: Same here:

> We could do the reverse we are doing in Datadog sink (maintaining a cumulative counter). I am leaving it for a future change?

sink/signalfx.go
```diff
@@ -136,8 +136,10 @@ func (s *SignalFx) Send(ctx context.Context, m *blip.Metrics) error {
 
 			// Convert Blip metric type to SFX metric type
 			switch metrics[i].Type {
-			case blip.COUNTER:
+			case blip.CUMULATIVE_COUNTER:
 				dp[n] = sfxclient.CumulativeF(name, dim, metrics[i].Value)
+			case blip.DELTA_COUNTER:
+				dp[n] = sfxclient.Counter(name, dim, int64(metrics[i].Value))
```
joycse06, line 142: `int64(float64)` is going to truncate the value (rather than rounding it), It probably doesn't matter in this case.

docs/v1.0/sinks/datadog.md
```diff
@@ -22,6 +22,15 @@ Only one of the methods can be used at a time.
 If `dogstatsd-host` option is set, DogStatsD is used for sending metrics, otherwise `api` keys must be provided.
 If the `doststatsd-host` doesn't include the port, `8125` is used as the default port.
 
+## Counter Metrics
+Previously, blip treated most counter metrics as cumulative counters during collection, except for those that were truncated after collection. Although certain metrics platforms support cumulative counters as a metric type, Datadog only supports delta as counter values.
+
```
Areson, line 27: I'm not sure this is strictly true. It supports cumulative counters, but it doesn't really support treating them as a rate though, whereas other platforms like SignalFX could/would.

metrics/aws.rds/rds.go
```diff
@@ -194,7 +194,7 @@ func (m *RDS) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 				},
 			}
 			if isCounter[metric] {
-				m.Type = blip.COUNTER
+				m.Type = blip.DELTA_COUNTER
```
daniel-nichter, line 197: Is this right? I'm not sure because different platforms differ on whether counter means cumulative or delta. In my mind, counter == cumulative.

metrics/status.global/global.go
```diff
@@ -20,10 +20,10 @@ import (
...
 
 /*
-	Begin the collector with a const block.
-	First define DOMAIN.
-	Then define any OPT_* const.
-	Then define other const as needed.
+Begin the collector with a const block.
+First define DOMAIN.
+Then define any OPT_* const.
+Then define other const as needed.
```
daniel-nichter, line 26: Did basic go fmt change this, or one of our Block git hooks?

prom/tr/generic.go
daniel-nichter: Afaik, Prom send cumulative counters. E.g. last time I looked at what msyqld_exporter did, it was raw numbers from MySQL (i.e. cumulative counters).

daniel-nichter: https://prometheus.io/docs/concepts/metric_types/

sink/signalfx.go
daniel-nichter: Yeah shouldn't matter because they're really integers from MySQL. Blip uses float64 because that's more flexible overall for monitoring. I.e. if a tool uses int and then needs float, it's not really possible. But if a tool uses float, that works for the rare floats and ints.

sink/datadog.go
```diff
@@ -126,7 +138,8 @@ func NewDatadog(monitorId string, opts, tags map[string]string, httpClient *http
 
 		case "api-compress":
 			d.compress = blip.Bool(v)
-
+		case "send-counter-delta":
+			d.sendCounterDelta = blip.Bool(v)
```
daniel-nichter, line 142: Why this option if only delta counter is valid for DD?

sink/datadog.go
```diff
@@ -276,9 +289,30 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
...
+					if !ok {
+						blip.Debug("error calculating delta for the counter: %s, no previous value found, skipping this run (next run will work)", name)
+						s.counters[metricID] = metrics[i].Value
+						continue METRICS
+					}
+					// previous value present, calculate delta
+					delta := metrics[i].Value - val
+					s.counters[metricID] = metrics[i].Value
+					if delta < 0 {
+						blip.Debug("error calculating delta for the counter: %s, got negative delta (can happen due to restart), skipping this run (next run will work).", name)
```
daniel-nichter, line 308: If negative due to restart, then probably we return the current value? Granted, that value is partial depending on where/how long into the restarted happened between collection, but partial is still correct and better then null.

sink/datadog.go
```diff
@@ -276,9 +289,30 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 
 			// Convert Blip metric type to Datadog metric type
 			switch metrics[i].Type {
-			case blip.COUNTER:
+			case blip.CUMULATIVE_COUNTER, blip.DELTA_COUNTER:
```
daniel-nichter, line 292: Seems like you need only `case blip. CUMULATIVE_COUNTER`? If it's already a delta, then no extra work is needed? (Or, if delta, then just `metricVal := metrics[i].Value`.)

metrics/wait.io.table/table.go
```diff
@@ -264,9 +264,13 @@ func (t *Table) Collect(ctx context.Context, levelName string) ([]blip.MetricVal
 		tblName = *values[1].(*string)
 
 		for i := 2; i < len(cols); i++ {
+			counterType := blip.CUMULATIVE_COUNTER
+			if o.truncate {
```
daniel-nichter, line 268: Can this logic be done once in `Prepare`?

sink/datadog.go
```diff
@@ -470,6 +504,41 @@ func (s *Datadog) estimateSize(metrics []datadogV2.MetricSeries) (int, error) {
 	return size / len(metrics), nil
 }
 
+func (s *Datadog) uniqIdentifierForMetric(name string, tags []string) string {
```
daniel-nichter, line 507: This and `intForTag` aren't clear even though I have a rough idea of what you're trying to do: a metrics is unique only as `name+all tags`. And since some collectors get metrics for a long list of objects (all tables, for example), there can be a long list of unique metrics, like:
* count_read,db1,tbl1
* count_read,db1,tbl2
* ...
* count_read,dbN,tblN

What I don't get is how mappping to uint16 helps or works? (And why uint16 specifically, and not unit or uint32? And if "short" is the goal, why `+= "_"`?)

I don't think we need to sort because Go maps return hash values in consistent order.  Even even despite that, isn't a unique metric possible to compute by doing something like SHA(name,tags...) regardless of tag order?

sink/datadog.go
Areson: I asked for this to be switchable in order to help with transitioning. Being able to take newer blip versions but not have this turned on yet would be helpful for that.

sink/datadog.go
Areson: A loss of metric would aid in detecting an issue though, if it persisted.

sink/datadog.go
Areson: I original proposed a version of this. The idea is that metrics are distinct on name + tags/tag values, and for some metrics we would have a large variation. Large strings result in slower comparisons and additional memory overhead, so the idea was that instead of making a really big string and/or additionally hashing it that we map `tag:value` to an int sequentially as a very lightweight pseudo-hash in order to produce a unique key with a relatively smaller string. The `"_"` is unnecessary. There is also a bug in this implementation that I'll call out in a separate comment.

sink/datadog.go
```diff
@@ -470,6 +504,41 @@ func (s *Datadog) estimateSize(metrics []datadogV2.MetricSeries) (int, error) {
...
+}
+
+// intForTag returns an uint16 for each tag so the uniq identifier for each metric and tags
+// combination we need to keep track of previous counter values are short.
+// it also maintains a separate counter for each tag key so the numbers don't get large
+func (s *Datadog) intForTag(tag string) uint16 {
+	v, ok := s.tagToInt[tag]
+	if !ok {
+		tokens := strings.Split(tag, ":")
+		next, ok := s.tagValues[tokens[0]]
```
Areson, line 527: I'm realizing that if we don't also include the _value_ for the tag someplace that this won't work properly, as we are creating a "unique " identifier on only the tag names and not the values.

Areson (changes requested):

sink/datadog.go
Areson: Would we really need to put a delta value in the map? It seems like if it is `DELTA_COUNTER` that we want to avoid adding it to the map to reduce memory overhead.

metrics/status.global/global.go
joycse06: Yes, it was go fmt change but from go 1.20 which I had set as the version in my goland. I have changed it back to 1.18 and it's not reformatting it anymore. So it's a future go fmt change.

Saying that, I think we should use hermit to mange and pin go version in blip and also upgrade to 1.20. I will do that in a seperate PR. I am going leave this change maybe as it's going to be reformatted like this in future version of go fmt anyway?

metrics/wait.io.table/table.go
joycse06: Sure, can do!

prom/tr/generic.go
joycse06: Yes, prom send Cumulative counters. As this was switching on Metric Types, I need to include `DELTA_COUNTER` as a case for completeness.

I am ignoring delta values for platforms which don't support them atm. But we could do reverse of what we are doing in Datadog sink to accumulate deltas.

sink/datadog.go
joycse06: I am not putting the `DELTA_COUNTER` in the map, I have only merged them into a single case as they have 15+ line in common which I will otherwise have to duplicate. But if it's confusing, I can make them seperate cases.

The extra calculation required for `CUMULATIVE_COUNTER`s is branched with 

```
if s.sendCounterDelta && metrics[i].Type == blip.CUMULATIVE_COUNTER {
```

The value is sent as is, if its a `DELTA_COUNTER`

sink/datadog.go
joycse06: A loss of metric would probably happen by not receiving the metric at all in the sink which will be detected by no data in DD.

I agree with Dan, we can send the partial value we will get after restart. I will update.

sink/datadog.go
joycse06: You are right, as for metrics with tablenames (or db) as tags would need them in the token to uniquely identify them.

We can just use the full `tag:value` pair rather than splitting it.

sink/datadog.go
joycse06: > Even even despite that, isn't a unique metric possible to compute by doing something like SHA(name,tags...) regardless of tag order?

Thats a good point. I think a sha256 sum of (name, tags...) might just be enough. We can go with a sha sum and only consider optimising further if we see any performance issues.

> And if "short" is the goal, why += "_"?
I added that before when I was putting in whole tag values in the hash so in case I had to debug I would see them separated in the log. I can remove it.

daniel-nichter: >Go maps return hash values in consistent order
I was so wrong: https://go.dev/ref/spec#For_statements and https://go.dev/doc/go1.3#map
I have a specific (but now faded) memory from long ago about this, but perhaps my memory got reversed over time: this changed (for small maps) way back in go1.3... or maybe I'm remember something else and the memory file has been corrupted to read as "map iteration."

metrics/status.global/global.go
daniel-nichter: Don't use Hermit on this or any of our open source projects. Hermit is a Cash thing (not even a Square thing). I know Hermit is also open source, but an open source project should stick to go mod and nothing else. If users want Hermit or something else, they wrap the open source project in one of their own internal projects.

sink/datadog.go
daniel-nichter: Why not stay on 1.0 and flip versions when ready? Because even if you upgrade to 1.1, since these vars aren't dynamic, you'll have to restart Blip to switch to the new, correct behavior (delta counters).

joycse06: Ok I will remove it in that case.

docs/v1.0/sinks/datadog.md
joycse06: If you send cumulative counter values to Datadog, even though you can do some calculations with the data (which are not dependant on metric type like `.rate(query)`, sending cumulative counter values datadog confuses their system and the in-app modifiers (`.as_count()` and `.as_rate()`) doesn't work as expected, they are enforced in their UIs (you can remove them from raw query editor but as soon as change something else from the visual query editor they come back).

One of their support rep told me if we wanted to keep the cumulative counter, we would be better off changing the type to `Guage` so their UI won't try to enforce methods which only works with Delta counters.

With all those issues, I think their support for Cumulative counters is confusing _at best_.

metrics/aws.rds/rds.go
joycse06: From the DD values they behaved liked delta counters. I will confirm with AWS support.

sink/datadog.go
```diff
@@ -292,9 +281,10 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
 			case blip.CUMULATIVE_COUNTER, blip.DELTA_COUNTER:
 				metricVal := metrics[i].Value
 
-				if s.sendCounterDelta && metrics[i].Type == blip.CUMULATIVE_COUNTER {
+				if metrics[i].Type == blip.CUMULATIVE_COUNTER {
 					// if the value is not specifically marked as DELTA, calculate delta
-					metricID := s.uniqIdentifierForMetric(name, tags)
+
+					metricID := s.uniqIdentifierForMetric(name, metrics[i].Group)
```
Areson, line 287: If we are going to use just `Group` and not `Meta` as well then we need to review some of the existing metrics as some of them use `Meta` instead of `Group` for tags that uniquely identify the metric values.

metrics/status.global/global.go
joycse06: I like `hermit` (or any tool for pinning tool versions), so you won't have to manually handle versions in different environments (dev, CI etc).

The problem with not using any such tool is exactly what happened here, everyone will have to explicitly remember to use the correct compiler every time they want to do something (IDE, cli). I wish go toolchain provided a way to enforce versions (will be hard to do in a sane way due to backward compatibility and all).

sink/datadog.go
joycse06: Discussed offline. We are going to exclude `meta` for now. The metrics which currently uses meta to uniquely identify values are `percona.qrt` and `query.response-time` which are both guages.

Dan mentioned: by design meta isn't supposed to identify, identity comes from domain + metric + group (if any).

sink/datadog.go
```diff
@@ -276,9 +278,31 @@ func (s *Datadog) Send(ctx context.Context, m *blip.Metrics) error {
...
 			// Convert Blip metric type to Datadog metric type
 			switch metrics[i].Type {
-			case blip.COUNTER:
+			case blip.CUMULATIVE_COUNTER, blip.DELTA_COUNTER:
+				metricVal := metrics[i].Value
+
+				if metrics[i].Type == blip.CUMULATIVE_COUNTER {
+					// if the value is not specifically marked as DELTA, calculate delta
+
+					metricID := s.uniqIdentifierForMetric(name, metrics[i].Group)
```
daniel-nichter, line 287: "Naming things" :-) Abbreviating "unique" to "uniq" is not standard because it's not much of an abbreviation when it's only off by 2: "ue". Plus abbreviating that but spelling out "identifier" (10 chars) instead of using "id" is weird (why save 2 but not 8?). And identifiers are unique by definition, so the two words together are redundant. And since Go likes short names "forMetric" isn't adding any value given the args and the value to which the return value is assigned.

Instead: `metricId := s.metricId(name,  metrics[i].Group))`.  And below:

```go
// metricId returns the metric name concatenated with sorted group keys.
// For example, if the metric name is "foo" and the group keys are "a" and "b",
// it returns "fooab". This is used to calculate delta counter values in Collect.
func (s *Datadog) metricId(name string, groups map[string]string) string {
```

sink/datadog.go
```diff
@@ -470,6 +494,35 @@ func (s *Datadog) estimateSize(metrics []datadogV2.MetricSeries) (int, error) {
...
+	sortedGroupValues := s.valuesInSortedKeyOrder(groups)
+
+	var key string
+	key += name
+	key += sortedGroupValues
+
+	return key
+}
+
+func (s *Datadog) valuesInSortedKeyOrder(tags map[string]string) string {
```
daniel-nichter, line 507: Don't split out functions when they're not used elsewhere. This func _is_  the bulk of `uniqIdentifierForMetric` (or `metricId` as I'd call it). This kind separation immediately raises the question (to future devs): "Is  valuesInSortedKeyOrder used elsewhere? or other things?" But no, it's not.

daniel-nichter (changes requested):

metrics/wait.io.table/table.go
```diff
@@ -197,6 +198,7 @@ LEVEL:
 		} else {
 			o.truncateTimeout = 250 * time.Millisecond // default
 		}
+		o.metricType = blip.CUMULATIVE_COUNTER
```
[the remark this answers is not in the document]
joycse06: I think it's the opposite, if you truncate after each collection you are going to get Delta on the next collection. I could move these into the If Block above to make it clearer, but it will work as is as well, because putting `o.metricType = blip.CUMULATIVE_COUNTER` here is equivalent to `if o.truncate == false` as I override it with Delta inside `if o.truncate` right after this assignment. But I will move anyway to make it clearer.

sink/datadog.go
joycse06: Refactored.

sink/datadog.go
joycse06: Inlined 👍

metrics/aws.rds/rds.go
joycse06: I only did this because we currently don't have any cumulative counters for RDS. I can main convert it into a map of maps (for both delta and cumulative counters) but cumulative counters will be empty anyway, but the code will be unnecessarily more complex (I will have to add code for checking if a metric is cumulative while knowing they never are atm.

I have added a comment to make it clearer, but if you still think making it complete will be better I will be happy to do so.

joycse06: The alternative is doing something similar to:

```go
var isCounter = map[byte]map[string]bool{
	blip.DELTA_COUNTER: {
		"AbortedClients":       true,
		"BacktrackWindowAlert": true,
	},
	blip.CUMULATIVE_COUNTER: {},
}
```

And convert the check to:

```go
if isCounter[blip.DELTA_COUNTER][metric] {
	m.Type = blip.DELTA_COUNTER
}

if isCounter[blip.CUMULATIVE_COUNTER][metric] {
	m.Type = blip.CUMULATIVE_COUNTER
}
```

daniel-nichter: Ack, comment is good; it's sufficient to help future eng avoid confusion/know what we intended to do today.

daniel-nichter (approved):

PR #116, opened 2023-12-18
Make HA as configurable plugin by Blip users

prudhvi:
Make the HA Manager customizable by the users of Blip.

For Cash app multiple clusters it's more appropriate to do leader election via k8s configmap so doing this lets us choose which EKS cluster blip is enabled/authorized to collect and send metrics, when blip is running in multiple EKS clusters.


ha/factory.go
```diff
@@ -0,0 +1,51 @@
+package ha
+
+import (
+	"fmt"
+	"github.com/cashapp/blip"
+	"sync"
+)
+
+type HAManagerFactory interface {
```
daniel-nichter, line 9: Remove "HA" prefix. In Go, pkg name is considered part of the full name context, so `ha.ManagerFactory` is the idiomatic style.

monitor/loader.go
```diff
@@ -7,6 +7,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"github.com/cashapp/blip/ha"
```
daniel-nichter, line 10: Sort the imports:
```
built-in

others
```

monitor/loader.go
```diff
@@ -463,11 +464,19 @@ func (ml *Loader) makeMonitor(cfg blip.ConfigMonitor) (*Monitor, error) {
 		sinks = append(sinks, sink)
 	}
 
+	// Configure the HA Manager for the monitor
+	var ham ha.Manager
```
daniel-nichter, line 468: "ham" 😆  This component will now officially be known as "the ham".

ha/factory.go
```diff
@@ -0,0 +1,51 @@
+package ha
```
daniel-nichter, line 1: Let's put this all in ha/ha.go since it's all closely related--one less file.

ha/factory.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+type HAManagerFactory interface {
+	Make(monitor blip.ConfigMonitor) (Manager, error)
+}
+
+type defaultFactory struct {
+}
+
+var df = &defaultFactory{}
```
daniel-nichter, line 16: In keeping with other pkgs (iirc):

```go
// Internal factory disabled by default (HA not used).
var f = Disabled
```

Then put the factor `Make` method on `disabled` (so it implements both `Manager` and `ManagerFactory` interfaces).

Then in `Register`, use `sync.Once` like you did in an earlier rev to avoid having locks and such.

daniel-nichter (approved):

PR #118, opened 2024-03-11
Update repl.lag collector to add option for collecting lag using performance schema

joycse06:
This follows the same pattern `var.global` uses to add options for collecting replication lag from performance schema.

I haven't updated the doc yet, as it seems current `repl.log` collector usage is documented in `Heartbeat` section, it probably doesn't make sense to add the `pfs` option in that page. So we have to restructure the doc a bit to document this update.

### DIFF
--- a/metrics/repl.lag/lag.go
+++ b/metrics/repl.lag/lag.go
@@ -141,6 +141,7 @@ LEVEL:
 		c.dropNotAReplica[levelName] = !blip.Bool(dom.Options[OPT_REPORT_NOT_A_REPLICA])
 		c.replCheck = sqlutil.CleanObjectName(dom.Options[OPT_REPL_CHECK]) // @todo sanitize better
 
+		fmt.Printf("Level: %s, Lag Writer: %q", levelName, dom.Options[OPT_WRITER])
 		switch dom.Options[OPT_WRITER] {
 		case LAG_WRITER_PFS:
 			c.lagWriterIn[levelName] = LAG_WRITER_PFS
--- a/metrics/repl.lag/lag.go
+++ b/metrics/repl.lag/lag.go
@@ -179,7 +179,7 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 		return c.collectPFS(ctx, levelName)
 	}
 
-	panic(fmt.Sprintf("invalid lag writer in Collect %q", c.lagWriterIn[levelName]))
+	panic(fmt.Sprintf("invalid lag writer in Collect %q. All levels: %v", c.lagWriterIn[levelName], c.lagWriterIn))
 }
 
 // //////////////////////////////////////////////////////////////////////////
--- a/metrics/repl.lag/lag.go
+++ b/metrics/repl.lag/lag.go
@@ -32,11 +32,26 @@ const (
 	// MySQL8LagQuery is the query to calculate approximate lag
 	// from replication worker stats in performance schema
 	// This is only available in MySQL 8 (and above) and when performance schema is enabled
-	MySQL8LagQuery = `SELECT TIMESTAMPDIFF(MICROSECOND,
-max(LAST_APPLIED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP),
-max(LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP)
-)/1000
-FROM performance_schema.replication_applier_status_by_worker`
+	MySQL8LagQuery = `WITH applier_latency AS (
+ SELECT TIMESTAMPDIFF(MICROSECOND, LAST_APPLIED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP, LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP)/1000 as applier_latency_ms
+ FROM performance_schema.replication_applier_status_by_worker ORDER BY LAST_APPLIED_TRANSACTION_END_APPLY_TIMESTAMP DESC LIMIT 1
+), queue_latency AS (
+ SELECT MIN(
+ CASE
+  WHEN
+   LAST_QUEUED_TRANSACTION = 'ANONYMOUS' OR
+   LAST_APPLIED_TRANSACTION = 'ANONYMOUS' OR
+   GTID_SUBTRACT(LAST_QUEUED_TRANSACTION, LAST_APPLIED_TRANSACTION) = ''
+  THEN 0
+   ELSE
+   TIMESTAMPDIFF(MICROSECOND, LAST_APPLIED_TRANSACTION_IMMEDIATE_COMMIT_TIMESTAMP, NOW(3))/1000
+ END
+) AS queue_latency_ms,
+IF(MIN(TIMESTAMPDIFF(MINUTE, LAST_QUEUED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP, NOW()))>1,'IDLE','ACTIVE') as queue_status
+FROM performance_schema.replication_applier_status_by_worker w
+JOIN performance_schema.replication_connection_status s ON s.channel_name = w.channel_name
+)
+SELECT IF(queue_status='IDLE',0,GREATEST(applier_latency_ms, queue_latency_ms)) as lagMs FROM applier_latency, queue_latency;`
 )
 
 type Lag struct {
@@ -165,7 +180,7 @@ LEVEL:
 		switch writer {
 		case LAG_WRITER_PFS:
 			// Try collecting, discard metrics
-			if _, err = c.collectPFS(ctx, levelName); err != nil {
+			if _, err = c.collectPFSv2(ctx, levelName); err != nil {
 				return nil, err
 			}
 		case LAG_WRITER_BLIP:
@@ -175,7 +190,7 @@ LEVEL:
 			}
 		case "auto", "": // default
 			// Try PFS first
-			if _, err = c.collectPFS(ctx, levelName); err == nil {
+			if _, err = c.collectPFSv2(ctx, levelName); err == nil {
 				blip.Debug("repl.lag auto-detected PFS")
 				writer = LAG_WRITER_PFS
 			} else {
@@ -205,7 +220,7 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 	case LAG_WRITER_BLIP:
 		return c.collectBlip(ctx, levelName)
 	case LAG_WRITER_PFS:
-		return c.collectPFS(ctx, levelName)
+		return c.collectPFSv2(ctx, levelName)
 	}
 
 	panic(fmt.Sprintf("invalid lag writer in Collect %q in level %q. All levels: %v", c.lagWriterIn[levelName], levelName, c.lagWriterIn))
--- a/metrics/repl.lag/lag.go
+++ b/metrics/repl.lag/lag.go
@@ -194,7 +194,7 @@ LEVEL:
 		switch writer {
 		case LAG_WRITER_PFS:
 			// Try collecting, discard metrics
-			if _, err = c.collectPFSv2(ctx, levelName); err != nil {
+			if _, err = c.collectPFS(ctx, levelName); err != nil {
 				return nil, err
 			}
 		case LAG_WRITER_BLIP:
@@ -204,7 +204,7 @@ LEVEL:
 			}
 		case "auto", "": // default
 			// Try PFS first
-			if _, err = c.collectPFSv2(ctx, levelName); err == nil {
+			if _, err = c.collectPFS(ctx, levelName); err == nil {
 				blip.Debug("repl.lag auto-detected PFS")
 				writer = LAG_WRITER_PFS
 			} else {
@@ -235,7 +235,7 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 	case LAG_WRITER_BLIP:
 		return c.collectBlip(ctx, levelName)
 	case LAG_WRITER_PFS:
-		return c.collectPFSv2(ctx, levelName)
+		return c.collectPFS(ctx, levelName)
 	}
 
 	panic(fmt.Sprintf("invalid lag writer in Collect %q in level %q. All levels: %v", c.lagWriterIn[levelName], levelName, c.lagWriterIn))
@@ -306,62 +306,6 @@ func (c *Lag) collectBlip(ctx context.Context, levelName string) ([]blip.MetricV
 		Type:  blip.GAUGE,
 		Value: float64(lag.Milliseconds),
 		Meta:  map[string]string{"source": lag.SourceId},
-	}
-	return []blip.MetricValue{m}, nil
-}
-
-func (c *Lag) collectPFS(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
-	var defaultLag []blip.MetricValue
-	if c.dropNotAReplica[levelName] {
-		defaultLag = nil
-	} else {
-		// send -1 for lag
-		m := blip.MetricValue{
-			Name:  "current",
-			Type:  blip.GAUGE,
-			Value: float64(-1),
-		}
-		defaultLag = []blip.MetricValue{m}
-	}
-
-	// if isReplCheck is supplied, check if it's a replica
-	isRepl := 1
-	if c.replCheck != "" {
-		query := "SELECT @@" + c.replCheck
-		if err := c.db.QueryRowContext(ctx, query).Scan(&isRepl); err != nil {
-			return nil, fmt.Errorf("checking if instance is replica failed, please check value of %s. Err: %s", OPT_REPL_CHECK, err.Error())
-		}
-	}
-
-	if isRepl == 0 {
-		return defaultLag, nil
-	}
-
-	// instance is a replica or replCheck is not set
-	var lagValue sql.NullString
-	if err := c.db.QueryRowContext(ctx, MySQL8LagQuery).Scan(&lagValue); err != nil {
-		return nil, fmt.Errorf("could not check replication lag, check that this is a MySQL 8.0 replica, and that performance_schema is enabled. Err: %s", err.Error())
-	}
-	if !lagValue.Valid {
-		// required performance schema table exists, otherwise the query would have returned error
-
-		// if replCheck is empty, we can assume based on the query that it's not a replica and return nil or -1
-		if c.replCheck == "" {
-			return defaultLag, nil
-		} else {
-			// it's a replica, so lagValue should be valid, but it's not so raise error
-			return nil, fmt.Errorf("cannot determine replica lag because performance_schema.replication_applier_status_by_worker returned an invalid value: %q (expected a positive integer value)", lagValue.String)
-		}
-	}
-
-	f, ok := sqlutil.Float64(lagValue.String)
-	if !ok {
-		return nil, fmt.Errorf("couldn't convert replica lag from performance schema into float. Lag: %s", lagValue.String)
-	}
-	m := blip.MetricValue{
-		Name:  "current",
-		Type:  blip.GAUGE,
-		Value: f,
 	}
 	return []blip.MetricValue{m}, nil
 }
--- a/metrics/repl.lag/lag.go
+++ b/metrics/repl.lag/lag.go
@@ -5,6 +5,7 @@ package repllag
 import (
 	"context"
 	"database/sql"
+	"fmt"
 	"strconv"
 	"time"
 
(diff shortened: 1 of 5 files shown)


metrics/repl.lag/lag.go
```diff
@@ -54,28 +68,29 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+				Name:    OPT_LAG_SOURCE,
+				Desc:    "How to collect Lag from",
+				Default: "auto",
 				Values: map[string]string{
-					"blip": "Native Blip replication lag",
-					//"pfs":    "MySQL Performance Schema",
+					"auto": "Auto-determine best source",
+					"blip": "Native Blip heartbeat replication lag",
+					"pfs":  "MySQL8 Performance Schema",
 					///"legacy": "Second_Behind_Slave|Replica from SHOW SHOW|REPLICA STATUS",
```
joycse06, line 79: I am still skipping this as I am not sure if it's going to be useful given the other options. But I can implement it as well for the sake of completeness.

metrics/repl.lag/lag.go
```diff
@@ -198,3 +265,58 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
...
+			m := blip.MetricValue{
+				Name:  "current",
+				Type:  blip.GAUGE,
+				Value: float64(-1),
+			}
+			return []blip.MetricValue{m}, nil
+		}
+	}
+	isRepl := 1
+	if c.replCheck != "" {
```
joycse06, line 285: I am not sure if we can skip this and just depend on the PFS query to determine if it's a replica.

test/docker/docker-compose.yaml
```diff
@@ -3,6 +3,7 @@ name: blip
 services:
     mysql57:
         image: mysql:5.7.34
+        platform: linux/amd64
```
joycse06, line 6: docker pull wasn't working in my M2 without these tags. Was getting the following error:

```
no matching manifest for linux/arm64/v8 in the manifest list entries
```

Adding the platform fixed it.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+	return nil, fmt.Errorf("auto lag source failed, last error: %s", err)
+}
 
-		c.dropNoHeartbeat[level.Name] = !blip.Bool(dom.Options[OPT_REPORT_NO_HEARTBEAT])
-		c.dropNotAReplica[level.Name] = !blip.Bool(dom.Options[OPT_REPORT_NOT_A_REPLICA])
+func (c *Lag) prepareBlipHeartbeatLag(levelName string, monitorID string, planName string, options map[string]string) (func(), error) {
+	if c.lagReader == nil {
+		c.dropNoHeartbeat[levelName] = !blip.Bool(options[OPT_REPORT_NO_HEARTBEAT])
 
+		table := options[OPT_HEARTBEAT_TABLE]
```
allanliu, line 200: Nit: If you're testing for map membership, it's more idiomatic to do this:
```
table, ok := options[OPT_HEARTBEAT_TABLE]
if !ok {
     table = blip.DEFAULT_HEARTBEAT_TABLE
}
```

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+			Table:      table,
+			SourceId:   options[OPT_HEARTBEAT_SOURCE_ID],
+			SourceRole: options[OPT_HEARTBEAT_SOURCE_ROLE],
+			ReplCheck:  c.replCheck,
+			Waiter: heartbeat.SlowFastWaiter{
+				MonitorId:      monitorID,
+				NetworkLatency: netLatency,
+			},
+		})
+		go c.lagReader.Start()
```
allanliu, line 226: no need to signal to the `Start` routine for a clean exit?

metrics/repl.lag/lag.go
```diff
@@ -198,3 +265,58 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 	}
 	return []blip.MetricValue{m}, nil
 }
+
+func (c *Lag) collectLagFromPFS(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	// if isReplCheck is supplied, check if it's a replica
+	defaultLag := func() ([]blip.MetricValue, error) {
```
allanliu, line 271: Is there a reason this needs to be an anonymous function definition? Can you just break it into a standard function since it seems like you're just justing the same default lag value? It also doesn't ever return an error so the signature is a bit misleading.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+// Prepare prepares lag collectors for all levels in the plan that contain the "repl.lag" domain
 func (c *Lag) Prepare(ctx context.Context, plan blip.Plan) (func(), error) {
 LEVEL:
-	for _, level := range plan.Levels {
+	for levelName, level := range plan.Levels {
 		dom, ok := level.Collect[DOMAIN]
 		if !ok {
 			continue LEVEL // not collected in this level
 		}
+		cleanup, err := c.prepareLevel(levelName, plan.MonitorId, plan.Name, dom.Options)
```
daniel-nichter, line 141: Inline `prepareLevel` here because its logic _is_ `Prepare`. Just blogged about this: https://hackmysql.com/post/fallacy-of-small-functions/.  Factoring out `prepareBlipHeartbeatLag` is probably ok since it actually has some "depth", but `preparePFS` can probably be inlined too: it's 2 lines of code, not counting the `return`.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
-			table = blip.DEFAULT_HEARTBEAT_TABLE
+func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	switch c.lagSourceIn[levelName] {
+	case LAG_SOURCE_BLIP:
+		return c.collectBlipHeartBeatLag(ctx, levelName)
+	case LAG_SOURCE_PFS:
+		return c.collectLagFromPFS(ctx, levelName)
+	}
+
+	panic(fmt.Sprintf("invalid lag source in Collect %s", c.lagSourceIn[levelName]))
```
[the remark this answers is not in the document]
daniel-nichter: In general I agree: panic as last resort. But I was think about this during the engine v2 rewrite and decided that Collect funcs should panic: https://github.com/cashapp/blip/blob/engine-v2/monitor/engine.go#L270. Reason being: if LCO and Engine are working correctly, they should never be called "out of level" like this. So if they are, it's not just an error, level collection is fundamentally messed up, which is panic-worthy.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+		return cleanup, nil
+	}
+
+	return nil, fmt.Errorf("auto lag source failed, last error: %s", err)
+}
 
-		c.dropNoHeartbeat[level.Name] = !blip.Bool(dom.Options[OPT_REPORT_NO_HEARTBEAT])
-		c.dropNotAReplica[level.Name] = !blip.Bool(dom.Options[OPT_REPORT_NOT_A_REPLICA])
+func (c *Lag) prepareBlipHeartbeatLag(levelName string, monitorID string, planName string, options map[string]string) (func(), error) {
+	if c.lagReader == nil {
```
daniel-nichter, line 197: Moving this block from a `select-case` to this func makes it need a little rewrite because now the bulk of the func is an uncommented `if` block (the comment got pushed down below) which means the normal code flow is not "reading down the left", as Go likes to do. So if we keep this func, something like:

```go
func (c *Lag) prepareBlip(...) ... {
    // Only need 1 lag read for all levels, so if it's already set just return early
    if c.lagReader != nil {
        return nil, nil
    }

    // Create 1 lag reader for all levels
    <most of code here>
}
```

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
-	}
+func (c *Lag) preparePFS(levelName string) error {
+	c.lagSourceIn[levelName] = LAG_SOURCE_PFS
+
+	// Try collecting, discard metrics
+	_, err := c.collectLagFromPFS(context.TODO(), levelName)
+	return err
+}
+
+func (c *Lag) collectBlipHeartBeatLag(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
```
daniel-nichter, line 248: We can shorten all these func names and make them consistent (e.g. this one doesn't have "From" but the other one does), presuming they're not inlined:
* prepareBlip
* collectBlip
* preparePFS
* collectPFS

metrics/repl.lag/lag.go
daniel-nichter: I don't know; try it out. Iirc this is mostly due to Aurora readers kind of looking like replicas but not.

metrics/repl.lag/lag.go
```diff
@@ -198,3 +265,58 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
 	}
 	return []blip.MetricValue{m}, nil
 }
+
+func (c *Lag) collectLagFromPFS(ctx context.Context, levelName string) ([]blip.MetricValue, error) {
+	// if isReplCheck is supplied, check if it's a replica
```
daniel-nichter, line 270: I think this comment got misplaced?

daniel-nichter (changes requested):

metrics/repl.lag/lag.go
joycse06: You are right, you can't tell if it's a reader or the cross-region writer of the replica from the output of the replica lag query from PF. So I will keep it around.

metrics/repl.lag/lag.go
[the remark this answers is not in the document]
joycse06: yeah, we do this in all other collectors as well.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+	}
+
+	panic(fmt.Sprintf("invalid lag source in Collect %s", c.lagSourceIn[levelName]))
+}
+
+// //////////////////////////////////////////////////////////////////////////
+// Internal methods
+// //////////////////////////////////////////////////////////////////////////
+
+func (c *Lag) prepareLevel(levelName string, monitorID string, monitorName string, options map[string]string) (func(), error) {
```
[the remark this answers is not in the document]
joycse06: Hmm, I know Dan usually recommend against using named returns, so I have been avoiding it. But I can see how this can help with readability. But to be consistent with other parts of code base I would like to hear @daniel-nichter 's thoughts before going ahead with this change.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,122 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+			default:
+				return nil, fmt.Errorf("invalid lag source: %s; valid values: auto, pfs, blip", source)
+			}
 		}
+	}
+
+	// -------------------------------------------------------------------------
+	// Auto source (default)
+	// -------------------------------------------------------------------------
+	var err error
```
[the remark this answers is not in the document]
joycse06: I am not just using it in the if conditional, the return also uses it, so need it outside of the scope of the if.

metrics/repl.lag/lag.go
joycse06: `c.lagReader.Stop()` is explicitly called in the cleanup func to stop the reader before exit.

metrics/repl.lag/lag.go
joycse06: You are right, it can just be a value. Doesn't need to be a func, I was probably anticipating it doing more when I wrote the func, but I don't need it to be a func.

metrics/repl.lag/lag.go
joycse06: Will refactor. I have read the blog post earlier, nice read!

metrics/repl.lag/lag.go
joycse06: Yes, will move.

metrics/repl.lag/lag.go
joycse06: In this instance, I want to set it to default if it's not set or it's **intentionally** set to "" (empty string), so I think this is cleaner as I don't need to distinguish between empty or unset key.

joycse06:
Addressed the feedback, PTAL 🙏

metrics/repl.lag/lag.go
```diff
@@ -54,28 +68,29 @@ func (c *Lag) Help() blip.CollectorHelp {
 		Domain:      DOMAIN,
 		Description: "Replication lag",
 		Options: map[string]blip.CollectorHelpOption{
-			OPT_WRITER: {
-				Name:    OPT_WRITER,
-				Desc:    "Type of heartbeat writer",
-				Default: "blip",
+			OPT_LAG_SOURCE: {
```
daniel-nichter, line 71: Let's not change the option because it's backwards-incompatible. `writer = blip|pfs` works because Perf Schema is the writer/where heartbeats are written.

metrics/repl.lag/lag.go
```diff
@@ -198,3 +259,58 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
...
+		return nil, fmt.Errorf("could not check replication lag, check that this is a MySQL 8.0 replica, and that performance_schema is enabled. Err: %s", err.Error())
+	}
+	if !lagValue.Valid {
+		// it is a MySQL 8 instance and performance schema is enabled, otherwise the query would have returned error
+		// if replCheck is empty, we can assume based on the query that it's not a replica and return nil or -1
+		if c.replCheck == "" {
+			return defaultLag, nil
+		} else {
+			// it's a replica, so lagValue should be valid, raise error
+			return nil, fmt.Errorf("instance is a MySQL 8 replica and performance schema is enabled, still could not calculate lag, please check manually. Lag: %q", lagValue.String)
```
daniel-nichter, line 302: To make error message more clear (because user can't/won't check manually): "cannot determine replica lag because performance_schema.replication_applier_status_by_worker returned an invalid value: %v (expected a positive integer value)"

daniel-nichter (changes requested):

metrics/repl.lag/lag.go
joycse06: I have also prefixed the heartbeat related options with `heartbeat-` which are not backward compatible as well. Maybe I should revert them too? The options are:

```
source-id
source-role
table
```

I prefixed them because they have no relevance when the using PFS.

metrics/repl.lag/lag.go
```diff
@@ -54,28 +68,29 @@ func (c *Lag) Help() blip.CollectorHelp {
...
+			OPT_LAG_SOURCE: {
+				Name:    OPT_LAG_SOURCE,
+				Desc:    "How to collect Lag from",
+				Default: "auto",
 				Values: map[string]string{
-					"blip": "Native Blip replication lag",
-					//"pfs":    "MySQL Performance Schema",
+					"auto": "Auto-determine best source",
+					"blip": "Native Blip heartbeat replication lag",
+					"pfs":  "MySQL8 Performance Schema",
```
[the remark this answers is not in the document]
joycse06: We are using the `replication_applier_status_by_worker` table which seems to be present in 5.7 as well ([doc](https://dev.mysql.com/doc/mysql-perfschema-excerpt/5.7/en/performance-schema-replication-applier-status-by-worker-table.html)) so I think `Performance Schema` is enough.

metrics/repl.lag/lag.go
joycse06: I will revert them other changes as well, though they are bit ambiguous without context of heartbeat writer, documentation should fill that context.

metrics/repl.lag/lag.go
```diff
@@ -115,70 +130,116 @@ func (c *Lag) Help() blip.CollectorHelp {
...
 		}
+		c.dropNotAReplica[levelName] = !blip.Bool(dom.Options[OPT_REPORT_NOT_A_REPLICA])
+		c.replCheck = sqlutil.CleanObjectName(dom.Options[OPT_REPL_CHECK]) // @todo sanitize better
 
-		table := dom.Options[OPT_TABLE]
-		if table == "" {
-			table = blip.DEFAULT_HEARTBEAT_TABLE
+		if source, ok := dom.Options[OPT_LAG_SOURCE]; ok {
+			if len(source) > 0 && source != "auto" {
+				switch source {
```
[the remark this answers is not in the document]
joycse06: Makes sense, I will update. We use [a similar form](https://github.com/cashapp/blip/blob/29f6a8032581516c7359364c4df27f858e47910a/metrics/var.global/global.go#L129-L149) (I based this off that) to this in `var.global` collector, we should refactor that as well when we touch it next time.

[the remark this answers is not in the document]
joycse06: Refactored it, it has different behavior though.

Previously, tests were passing with just:

```
    repl.lag:
      options: {}
```
I didn't have to explicitly, set `writer: auto`. Even though, `if source, ok := dom.Options[OPT_LAG_SOURCE]; ok {` was returning `false` (I am not sure if it's a test only problem or it's how options are parsed, I am not too familiar with options parsing), as the default case was outside, it was still setting the correct source.

Now, without setting `writer: auto` it fails in collection because the that condition failing and default of `auto` is not being handled. But we were still returning `nil, nil` signalling everything is alright.

I check the implementation of [var.global](https://github.com/cashapp/blip/blob/29f6a8032581516c7359364c4df27f858e47910a/metrics/var.global/global.go#L147-L165) which seems to be handling default case separately as well, so maybe that's by design?

metrics/repl.lag/lag.go
```diff
@@ -198,3 +259,58 @@ func (c *Lag) Collect(ctx context.Context, levelName string) ([]blip.MetricValue
...
+	var lagValue sql.NullString
+	if err := c.db.QueryRowContext(ctx, MySQL8LagQuery).Scan(&lagValue); err != nil {
+		return nil, fmt.Errorf("could not check replication lag, check that this is a MySQL 8.0 replica, and that performance_schema is enabled. Err: %s", err.Error())
+	}
+	if !lagValue.Valid {
+		// it is a MySQL 8 instance and performance schema is enabled, otherwise the query would have returned error
+		// if replCheck is empty, we can assume based on the query that it's not a replica and return nil or -1
+		if c.replCheck == "" {
+			return defaultLag, nil
+		} else {
```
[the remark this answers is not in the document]
joycse06: This is serving a different case not covered by previous `if {` return where:
1. `replCheck` is not set, means we don't know for sure if it's a replica or not
2.  `lagValue` will only be invalid if it's `null` which is the case on non-replicas. So if it's invalid and
    a. `replCheck` is `""`: We return default as this is not a replica
    b. `replCheck` is **set**, and we are still at this point in code (`replCheck` didn't trigger early return), so either `replCheck` is incorrect, or something else is wrong, and we should return an error.
    
I hope that makes sense.

metrics/repl.lag/lag.go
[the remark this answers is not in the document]
joycse06: There's panic in many other collectors, we should probably do a sweep after the engine v2 lands to make it consistent across collectors.

joycse06:
@daniel-nichter  PTAL. I think I have addressed all your feedbacks. 🙏

metrics/repl.lag/lag.go
```diff
@@ -16,32 +17,65 @@ import (
...
+	OPT_RENAME_DEFAULT_REPLICATION_CHANNEL = "rename-default-replication-channel"
+	OPT_NETWORK_LATENCY                    = "network-latency"
+
+	LAG_WRITER_BLIP = "blip"
+	LAG_WRITER_PFS  = "pfs"
+
+	// MySQL8LagQuery is the query to calculate approximate lag
+	// from replication worker stats in performance schema
+	// This is only available in MySQL 8 (and above) and when performance schema is enabled
+	MySQL8LagQuery = `WITH applier_latency AS (
```
daniel-nichter, line 36: This can be removed.

metrics/repl.lag/lag.go
```diff
@@ -16,32 +17,65 @@ import (
...
-	OPT_REPORT_NOT_A_REPLICA = "report-not-a-replica"
-	OPT_NETWORK_LATENCY      = "network-latency"
+	OPT_HEARTBEAT_SOURCE_ID                = "source-id"
+	OPT_HEARTBEAT_SOURCE_ROLE              = "source-role"
+	OPT_HEARTBEAT_TABLE                    = "table"
+	OPT_WRITER                             = "writer"
+	OPT_REPL_CHECK                         = "repl-check"
+	OPT_REPORT_NO_HEARTBEAT                = "report-no-heartbeat"
+	OPT_REPORT_NOT_A_REPLICA               = "report-not-a-replica"
+	OPT_RENAME_DEFAULT_REPLICATION_CHANNEL = "rename-default-replication-channel"
```
daniel-nichter, line 27: `default-channel-name` (default: ""), then Blip and MySQL line up: the default channel name is the same when not specified: an empty string.

metrics/repl.lag/lag.go
```diff
@@ -16,32 +17,65 @@ import (
...
+   TIMESTAMPDIFF(MICROSECOND, LAST_APPLIED_TRANSACTION_IMMEDIATE_COMMIT_TIMESTAMP, NOW(3))/1000
+ END
+) AS queue_latency_ms,
+IF(MIN(TIMESTAMPDIFF(MINUTE, LAST_QUEUED_TRANSACTION_ORIGINAL_COMMIT_TIMESTAMP, NOW()))>1,'IDLE','ACTIVE') as queue_status
+FROM performance_schema.replication_applier_status_by_worker w
+JOIN performance_schema.replication_connection_status s ON s.channel_name = w.channel_name
+)
+SELECT IF(queue_status='IDLE',0,GREATEST(applier_latency_ms, queue_latency_ms)) as lagMs FROM applier_latency, queue_latency;`
+
+	defaultChannelName = "default"
```
daniel-nichter, line 57: Empty string because Blip tries to mirror MySQL directly. It only changes stuff when MySQL is so odd that industry standard never uses the odd MySQL thing. But in this case, channel="" works because empty string is a valid and unique key in a `map[string]T`.

metrics/repl.lag/lag.go
daniel-nichter: Also, intention was to let user rename it whatever, not to force a new default of "default". E.g. a user might want to `default-channel-name: my-primary-db` or something.

metrics/repl.lag/pfs.go
```diff
@@ -0,0 +1,269 @@
+// Copyright 2024 Block, Inc.
```
daniel-nichter, line 1: Good use of separate file. :-) Iirc, we (or other team members) had this conversation. Usually less files is good, but when some chunk of code really stands alone (in this case, all this PFS logic really does), then a separate pkg files makes editing a little easier. 👍

metrics/repl.lag/pfs.go
```diff
@@ -0,0 +1,269 @@
...
+			return nil, fmt.Errorf("checking if instance is replica failed, please check value of %s. Err: %s", OPT_REPL_CHECK, err.Error())
+		}
+	}
+
+	if isRepl == 0 {
+		return defaultLag, nil
+	}
+
+	lastQueued := map[string]string{} // by replica/to relay log
+	lastProc := map[string]string{}   // by coordinator/to workers
```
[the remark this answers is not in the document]
joycse06: I have gone with `pfsLag`

metrics/repl.lag/lag.go
joycse06: Removed.

daniel-nichter (approved):

PR #139, opened 2025-01-21
Fix `sqlutil.INList`

Areson:
`sqlutil.INList` repeats the first value of the list and leaves the last value off the resulting string. Fixed this bug and replaced uses of unescaped `IN` list creation with calls to `sqlutil.INList` where appropriate.

Adjusted metrics and heartbeat to use driver-level parameter interpolation instead of custom escaping, deprecating several function in `sqlutil`. Removed the use of `sqlutil.INList` in existing blip code but marked as deprecated.


daniel-nichter (approved):
We need to either redo this code or test it a lot better. Messing with quoting is a classic problem and risk (wrt injections).

PR #147, opened 2025-03-17
Add Detailed Debugging to query latency

Areson:


metrics/query.response-time/response_time.go
```diff
@@ -222,6 +222,27 @@ func (c *ResponseTime) Collect(ctx context.Context, levelName string) ([]blip.Me
...
 		}
 		metrics = append(metrics, m)
+
+		blip.Debug("[%s]: Formated percentile value: %s=%f", DOMAIN, percentile.formatted, us)
+	}
+
+	// If debugging is turned on dump the raw values from performance_schema.events_statements_histogram_global
+	if blip.Debugging {
+		blip.Debug("Bucket Number|Bucket Timer Low|Bucket Timer High|Count Bucket|Count Bucket and Lower|Bucket Quantile")
+		rows, err := c.db.QueryContext(ctx, "SELECT * FROM performance_schema.events_statements_histogram_global")
```
daniel-nichter, line 232: Probably best to order by what Blip will actually see/use, then also handle logging timestamps, i.e. the fact that logs will have virtually same ts means they'll order randomly in DD, so we won't be able to see the _real_ order of rows. Maybe emit just 1 big log event to bypass all log ordering (and possibly log filtering)?

Areson: I *think* that it will get split by newline even if I combine it, though that may fix the ordering issue. I'll test it out and see what happens.

Areson: Updated. Output from a single large string is broken up but the ordering is preserved.

daniel-nichter (approved):

PR #151, opened 2025-05-06
Adding the `error.*` domains

Areson:
Adding the `metrics/error` package that includes the following domains:
- `error.global`
- `error.account`
- `error.host`
- `error.thread`
- `error.user`

These domains emit metrics on errors from the error summary tables (https://dev.mysql.com/doc/refman/en/performance-schema-error-summary-tables.html) See the documentation for more details.

### DIFF
--- a/docs/content/metrics/domains/error.account/_index.md
+++ b/docs/content/metrics/domains/error.account/_index.md
@@ -1,0 +1,130 @@
+---
+title: "error.account"
+---
+
+The `error.account` domain includes server error metrics grouped by account from [error summary tables](https://dev.mysql.com/doc/refman/en/performance-schema-error-summary-tables.html).
+
+{{< toc >}}
+
+## Usage 
+
+For example:
+
+```
+mysql> SELECT * FROM performance_schema.events_errors_summary_by_account_by_error WHERE error_number = 3024 AND USER IN ('user')\G
+*************************** 1. row ***************************
+             USER: user
+             HOST: localhost
+     ERROR_NUMBER: 3024
+       ERROR_NAME: ER_QUERY_TIMEOUT
+        SQL_STATE: HY000
+ SUM_ERROR_RAISED: 1
+SUM_ERROR_HANDLED: 0
+       FIRST_SEEN: 2025-04-30 16:30:16
+        LAST_SEEN: 2025-04-30 16:30:16
+```
+
+The `SUM_ERR_RAISED` value is used for the Blip `raised` metric. `ERROR_NUMBER`, `ERROR_NAME`, `USER`, and `HOST` are used to group the metric.
+
+## Derived Metrics
+
+None.
+
+## Options
+
+### `all`
+
+|Value|Default|Description|
+|-----|-------|-----------|
+|yes  |&check; |Collect metrics on _all_ errors (not recommended)|
+|no   | |Collect only metrics for the errors listed in the plan|
+|exclude| |Collect metrics only for errors that are not listed in the plan|
+
+### `total`
+
+|Value|Default|Description|
+|-----|-------|-----------|
+|yes  |&check; |Return the total number of errors raised|
+|no   | |Do not return the total number of errors raised|
+|only| |Only return the total number of errors raised|
+
+If `all` is not set to `yes` the total will only reflect those errors that are specifically included or not excluded from collection.
+
+### `include`
+
+| | |
+|---|---|
+|**Value Type**|CSV string of accounts|
+|**Default**||
+
+A comma-separated list of ids to include. Overrides option `exclude`. The values for the filter should be formatted as `user@hostname`. Wildcards are allowed for either `user` or `hostname`, but not both. Invalid values will be ignored.
+
+### `exclude`
+
+| | |
+|---|---|
+|**Value Type**|CSV string of accounts|
+|**Default**||
+
+A comma-separated list of ids to exclude. Ignored if `include` is set. The values for the filter should be formatted as `user@hostname`. Wildcards are allowed for either `user` or `hostname`, but not both. Invalid values will be ignored.
+
+### `truncate-table`
+
+|Value|Default|Description|
+|---|---|---|
+|yes| |Truncate table after each successful collection|
+|no|&check; |Do not truncate table|
+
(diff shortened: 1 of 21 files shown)


daniel-nichter (changes requested):
There are 5 domains here:
* `error.global`
* `error.user`
* `error.thread`
* `error.host`
* `error.account`

We need to break them down as such because, for example, this collector can't collect global and user; it can only collect one. And by combining them, there are options that only make sense for one domain but not the others. (And overall it makes the collector code larger, more complicated than needed.)

Let's start with just `error.global` (i.e. events_errors_summary_global_by_error) with a default configuration that reports the running error count as-is (including zero) via metric `error.global.raised`. As a user, that's what I'd expect. (Deltas are a DD thing; the default is cumulative counter.)

If we need the other error domains/collectors, we can add them too--but I don't think we need them right now? But since you already have the code, if you want to add them too for completeness, that's fine.

daniel-nichter:
Re default:

In the case of size.table, for example, the obvious best default is per-table because experience tells us that's the normal question "How large are (each of) my tables?" And sometimes: "How large are my tables _in total_?", which is easy to aggregate in one's metric platform (Grafana, DD, whatever). -- Or to put it another way: if a monitor defaulted only to _total_ table size and required me to enable per-table size, I'd say "WTF?"

I don't think the same is true for errors. Instead, due to monitoring practices like RED and USE, the normal question is "Is my server spewing errors?" The user isn't normally interested in which errors because the ideal state is zero errors. (Unlike tables that are expected to have a size, so we want to know what those sizes are.) That's why a single sum (all error codes) is useful: if it's (near) zero, then great--I'm done. So unlike table sizes, I would _not_ say "WTF?" in this case. But...

But of course I agree that knowing specific errors is useful, too, even necessary. So really this is a product design question: what's the default setting/UX? I'm tempted to match the behaviour and options of table.size (e.g. option `total`, default "yes")... 🤔

Re metric type:

Cumulative counter is definitely correct. At the collector level, they (collectors) simply mirror MySQL, and in MySQL this is a (cumulative) counter. All the other collectors follow this. Delta, rate, and anything else is purely a metric sink and/or metric platform responsibility to modify as they need/wish. This is necessary (and has proven valuable) because there has to be some unknown, invariant, stable foundation--and this is it: collectors report stuff straight from MySQL _as is_.

daniel-nichter:
Re further discussions, here's what I think we need to do:

1. Break up this collector (separate domains, as listed above--5 of them)
2. Primary concern to start is `error.global`, so I'll just speak in terms of that one for now...
3. Remove `OPT_SUMMARIZE` because of 1 (not needed any longer)
4. Remove `OPT_COLLECT_EMPTY` because technically they're not "empty", they're zero
5. Since MySQL went crazy in 8.0.11 for reasons unknown, adding _tons_ of apparently empty "error messages", the global error table is (sadly) > 5 _thousand_ rows long, so collecting only `WHERE SUM_ERROR_RAISED > 0` is sensible and required
6. Options/defaults: total=yes (by default like other collectors) _and_ per-error code. (Given 5, this shouldn't yield too many error codes.)
7. Do the cumulative or delta counter same as io.wait.table. (For future ref: it's a cumulative counter in MySQL; this is correct. But when the TRUNCATE option is used it technically becomes a delta counter; a cumulative counter in this case would be incorrect. So we're not fiddling with the metric type or introducing an abstraction at this low level, we're making the type correct based on table truncation or not.)

Since the other domains are nearly identical, I'm sure there's a way to factor out common code, but that's an implementation detail. From the user point of view, there should be distinct domains/collectors.

metrics/error.global/error.go
```diff
@@ -0,0 +1,305 @@
...
+			},
+		},
+		Groups: []blip.CollectorKeyValue{
+			{Key: GRP_ERR_NUMBER, Value: "the error number, or an empty string for a total"},
+			{Key: GRP_ERR_NAME, Value: "the error name, or an empty string for a total"},
+		},
+		Metrics: []blip.CollectorMetric{
+			{
+				Name: "raised",
+				Type: blip.DELTA_COUNTER,
```
daniel-nichter, line 127: Since truncate=no by default, shouldn't this be cumulative counter by default?

metrics/error.global/error.go
```diff
@@ -0,0 +1,305 @@
...
+			}
+
+			errOpts.lockWaitQuery = fmt.Sprintf(LOCKWAIT_QUERY, int64(lockWaitTimeout))
+			errOpts.truncateErrPolicy = errors.NewTruncateErrorPolicy(dom.Errors[ERR_TRUNCATE_FAILED])
+			blip.Debug("error policy: %s=%s", ERR_TRUNCATE_FAILED, errOpts.truncateErrPolicy.Policy)
+		}
+
+		t.options[level.Name] = errOpts
+
+		// Run an initial truncate to clear out any old data
```
daniel-nichter, line 213: Not sure `Prepare` is right place for this... hard to say. Is truncate on startup a setup (prepare) thing? Or part of collection (i.e. a pre-step and requirement to `Collect`)? Hm...

metrics/error.global/error.go
```diff
@@ -0,0 +1,305 @@
...
+
+	var (
+		metrics   []blip.MetricValue
+		errors    int64
+		errorNum  string
+		errorName string
+		total     float64
+	)
+
+	// Are we truncating the table?
```
daniel-nichter, line 244: Comment looks wrong (copy-paste error?).

metrics/error.global/error.go
Areson: It is. This is a typo.

metrics/error.global/error.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (t *ErrorGlobal) truncate(ctx context.Context, levelName string) error {
+	o, ok := t.options[levelName]
+	if !ok {
+		return nil
+	}
+
+	conn, err := t.db.Conn(ctx)
+	if err == nil {
```
daniel-nichter, line 289: Go reads down the left, so `err != nil` standard boilerplate here.

Areson:
@daniel-nichter Updated with domains in a common package.

daniel-nichter (approved):

PR #152, opened 2025-05-06
Auto Increment Metric

Areson:
Introduces the `autoinc` domain that monitors the utilization of auto-increment columns in the database.

### DIFF
--- a/metrics/auto_increment/auto_increment.go
+++ b/metrics/auto_increment/auto_increment.go
@@ -1,0 +1,143 @@
+// Copyright 2024 Block, Inc.
+
+package autoincrement
+
+import (
+	"context"
+	"database/sql"
+
+	"github.com/cashapp/blip"
+)
+
+const (
+	DOMAIN = "autoinc"
+
+	OPT_EXCLUDE = "exclude"
+	OPT_INCLUDE = "include"
+)
+
+// AutoIncrement collects auto-increment utilization for the autoinc domain.
+// https://dev.mysql.com/doc/refman/8.0/en/sys-schema-auto-increment-columns.html
+type AutoIncrement struct {
+	db *sql.DB
+	// --
+	query  map[string]string
+	params map[string][]interface{}
+}
+
+// Verify collector implements blip.Collector interface.
+var _ blip.Collector = &AutoIncrement{}
+
+// NewAutoIncrement makes a new AutoIncrement collector,
+func NewAutoIncrement(db *sql.DB) *AutoIncrement {
+	return &AutoIncrement{
+		db:     db,
+		query:  map[string]string{},
+		params: map[string][]interface{}{},
+	}
+}
+
+// Domain returns the Blip metric domain name (DOMAIN const).
+func (t *AutoIncrement) Domain() string {
+	return DOMAIN
+}
+
+// Help returns the output for blip --print-domains.
+func (t *AutoIncrement) Help() blip.CollectorHelp {
+	return blip.CollectorHelp{
+		Domain:      DOMAIN,
+		Description: "Auto Increment Utilization",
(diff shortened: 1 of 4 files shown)


Areson:
@daniel-nichter When you get a chance please drop some comments so I can polish this up.

metrics/factory.go
```diff
@@ -8,6 +8,7 @@ import (
 	"sync"
 
 	"github.com/cashapp/blip"
+	autoincrement "github.com/cashapp/blip/metrics/auto_increment"
```
daniel-nichter, line 11: `s/auto.increment/autoinc/g`.  "Autoinc" is standard MySQL vernacular. Plus, standardizing on this avoids this problem (next comment)...

metrics/auto_increment/auto_increment.go
```diff
@@ -0,0 +1,143 @@
...
+				Name:    OPT_EXCLUDE,
+				Desc:    "Comma-separated list of database or table names to exclude (ignored if " + OPT_INCLUDE + " is set)",
+				Default: "mysql.*,information_schema.*,performance_schema.*,sys.*",
+			},
+		},
+		Groups: []blip.CollectorKeyValue{
+			{Key: "db", Value: "the database name for the corresponding auto increment value"},
+			{Key: "tbl", Value: "the table name for the corresponding auto increment value"},
+			{Key: "col", Value: "the column name for the corresponding auto increment value"},
+			{Key: "data_type", Value: "the data type of the column"},
```
daniel-nichter, line 65: Is this needed?

metrics/auto_increment/auto_increment.go
```diff
@@ -0,0 +1,143 @@
...
+		},
+		Metrics: []blip.CollectorMetric{
+			{
+				Name: "percent_used",
```
daniel-nichter, line 69: Naming things is hard... `usage`? 🤔

metrics/auto_increment/auto_increment.go
```diff
@@ -0,0 +1,143 @@
+// Copyright 2024 Block, Inc.
```
daniel-nichter, line 1: s/2024/2025/. You can probably update this in all files.

metrics/auto_increment/auto_increment.go
Areson: We care about the distinction between `int` and `bigint` for things like triage. Whether or not it is signed also informs us about possible solutions. In terms of monitoring we may want to alert on things that are not `unsigned` or are `int` instead of `bigint`. We could split those into separate tags if we wanted.

metrics/auto_increment/query.go
```diff
@@ -0,0 +1,53 @@
+// Copyright 2024 Block, Inc.
+
+package autoincrement
+
+import (
+	"strings"
+)
+
+func AutoIncrementQuery(set map[string]string) (string, []interface{}, error) {
+	query := "SELECT table_schema, table_name, column_name, data_type, auto_increment_ratio, is_unsigned FROM sys.schema_auto_increment_columns WHERE auto_increment_ratio IS NOT NULL"
```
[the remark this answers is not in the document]
Areson: We _could_ as it is just a view on `INFORMATION_SCHEMA`. This particular `sys` view has been around since at least 5.7 so the internal discussion was that we were probably safe to use it rather than rolling our own calculations.

metrics/auto_increment/auto_increment.go
Areson: I'll make a todo for this in a separate PR.

Areson:
@daniel-nichter Updates to most of the above items is done. Comments on one.

daniel-nichter (approved):

PR #155, opened 2025-05-15
Manually update 2 deps that the Dependabot PR#150 can't build

daniel-nichter:


Areson (approved):
This appears to work fine based on my integration testing.

PR #159, opened 2025-07-23
Retrieve username from secret in the hotswap driver function

JashLal:
Update to the hotswap credentials function. Changes the logic for the `password-secret` authentication mode. Retrieves a `username` key in addition to `password` and if found sets the mysql config.

### DIFF
--- a/aws/secret.go
+++ b/aws/secret.go
@@ -25,7 +25,23 @@ func NewSecret(name string, cfg aws.Config) Secret {
 	}
 }
 
-func (s Secret) Password(ctx context.Context) (string, error) {
+func (s Secret) Username(v map[string]interface{}) (string, error) {
+	username, ok := v["username"]
+	if !ok {
+		return "", fmt.Errorf("failed to retrieve value for key 'username'")
+	}
+	return username.(string), nil
+}
+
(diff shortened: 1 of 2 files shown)


aws/secret.go
```diff
@@ -25,29 +25,44 @@ func NewSecret(name string, cfg aws.Config) Secret {
...
-func (s Secret) Password(ctx context.Context) (string, error) {
+func (s Secret) Username(v map[string]interface{}) (string, error) {
+	username, ok := v["username"]
+	if !ok {
+		return "", fmt.Errorf("failed to retrieve value for key 'username'")
+	}
+	return username.(string), nil
+}
+
+func (s Secret) Password(v map[string]interface{}) (string, error) {
```
daniel-nichter, line 36: I don't think these 2 new method are needed; they're extremely shallow (see book _Philosophy of Software Design_). They can be put inline in the callback func (`return func(ctx context.Context) (Credentials, error) {`) so the reader doesn't have make a mental context switch. E.g. when I fist saw these 2 methods, I figured they served some kind of other purpose/did something else. It took me a moment to realize they're just fetching a value from a map that's passed into them--there's no additional logic or handling or anything else.

PR #164, opened 2026-06-03
Add AWS password secret parser plugin

Areson:
## Why
Blip currently supports only the default RDS JSON shape for AWS Secrets Manager password payloads, so embedders with custom secret formats need to replace the whole DB factory to customize credential parsing.

## What
- Add `Plugins.ParsePasswordSecret` for mapping raw `SecretString` or `SecretBinary` bytes onto `blip.Secret`
- Keep `aws.Secret.GetSecret` source-compatible and add `GetSecretPayload` for raw secret bytes
- Use the parser in the Secrets Manager credential reload path while preserving default `username`/`password` behavior
- Add parser and credential-path tests, and update AWS/config/integration docs

## References
Validated with:

`go test ./... -run 'TestDefaultPasswordSecretParser|TestPasswordSecretCredentialFunc|^$'`

### DIFF
--- a/blip.go
+++ b/blip.go
@@ -99,24 +99,25 @@ type SinkFactoryArgs struct {
 	Tags      map[string]string // config.monitor.tags
 }
 
-// Secret is a MySQL username/password parsed from a secret payload.
-type Secret struct {
+// DbCredentials are MySQL credentials parsed or loaded for a connection.
+type DbCredentials struct {
 	Username string
 	Password string
+	TLS      ConfigTLS
 }
 
-// PasswordSecretParser maps an AWS Secrets Manager payload to a Secret.
-// The Secret argument is pre-populated with config defaults and can be modified
+// PasswordSecretParser maps an AWS Secrets Manager payload to database credentials.
+// The DbCredentials argument is pre-populated with config defaults and can be modified
 // by the parser. The payload is the raw SecretString or SecretBinary value.
-type PasswordSecretParser func(context.Context, ConfigMonitor, []byte, *Secret) error
+type PasswordSecretParser func(context.Context, ConfigMonitor, []byte, *DbCredentials) error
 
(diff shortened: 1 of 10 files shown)


dbconn/factory.go
```diff
@@ -288,35 +304,7 @@ func (f factory) Credentials(cfg blip.ConfigMonitor) (CredentialFunc, error) {
 			return nil, err
 		}
 		secret := aws.NewSecret(cfg.AWS.PasswordSecret, awscfg)
```
daniel-nichter: I'd move this into `passwordSecretCredentialFunc` because I think that's the only place it's used, and that func is already passed `cfg` and (as a method on the factory) has `f.awsConfig` too.

Ok bye, back to the good life. ;-)

blip.go
```diff
@@ -98,6 +99,54 @@ type SinkFactoryArgs struct {
 	Tags      map[string]string // config.monitor.tags
 }
 
+// Secret is a MySQL username/password parsed from a secret payload.
+type Secret struct {
```
daniel-nichter: Maybe raise `dbconn.Credentials` to `blip.DbCredentials`? Reason being: 1) "Secret" is, here, actually just db creds, and 2) this Secret struct is really only used to plumb down to Credentials struct (as in the snippet below). So by raising the latter, you avoid plumbing and create a more self-describing user-facing struct.

```go
return Credentials{
	Password: parsedSecret.Password,
	Username: parsedSecret.Username,
}, nil
```

daniel-nichter (approved):
:stamp:

...

Wait a minute! I don't work for you any longer! :-D

/me goes back to chopping wood

Areson:
> /me goes back to chopping wood

Thanks for the stamp Hemingway!

JashLal (approved):

PR #175, opened 2026-08-07
Generalize external database modules

Areson:
## Why
Keep Blip purpose-built for MySQL while allowing pgblip and future database modules to reuse its monitor, plan, transform, and sink runtime without adding engine implementations to core.

## What
- Replace PostgreSQL-specific core configuration with an opaque registered database-module contract
- Generalize collector compatibility, factory delegation, and credential port handling
- Preserve MySQL defaults while adding strict module validation, redaction, and type-preserving config interpolation

## Risk Assessment
Low — the change targets an unreleased feature stack, leaves omitted database types on historical MySQL behavior, and passed full local unit, race, vet, and build validation.

## Stack
1. #176 `ioberst/bgblip-module` → `main`
2. #177 `ioberst/multi-database-monitor` → `ioberst/bgblip-module`
3. #174 `Areson/postgres-config-guards` → `ioberst/multi-database-monitor`
4. **This PR:** #175 `Areson/database-module-extensibility` → `Areson/postgres-config-guards`

Review bottom-to-top; merge in reverse order: #175, #174, #177, then #176. Keep the stack in draft until the manual Blippy PostgreSQL staging canary passes.

Generated with Codex


daniel-nichter:
🤔

Side note: making Blip work for pg crossed my mind at the beginning, but back then the chance of SQ needing it for pg was negligible (even considering AP), so the YAGNI argument won. Funny how things change.

Areson:
Well...this was pushed a litter sooner than I expected.

But yes, I'm working on it. I was torn about extending the "purpose build MySQL" tool to pg, so I'm currently taking the route of trying to generalize bits of the core around auth, etc. (which is pretty engine-agnostic) and allow extending it via external plugins. The end goal being that core `blip` stays focused on MySQL, but can load external DB engine modules to handle other flavors.

PR #176, opened 2026-08-07
Support external database modules

Areson:
## Why
Allow official external database modules to reuse Blip's monitor, plan, collection, transformation, and sink runtime while keeping Blip itself purpose-built for MySQL and preserving existing MySQL behavior.

## What
- Add opt-in database module registration, opaque module configuration with per-placeholder interpolation, and strict module-owned validation
- Add engine-neutral credential callbacks and composable database/provider factories, including monitor-owned multi-database providers and strict provider validation
- Add optional collector database compatibility metadata and validate plans against their monitors
- Reject MySQL-only subsystems for external monitors while retaining MySQL defaults for existing configurations and implementations
- Document the external module contract and cover its configuration, lifecycle, compatibility, and rollback paths

## Risk Assessment
Medium — this introduces public extension seams and adjusts monitor lifecycle internals, but Blip does not register or ship another database engine. Existing configurations, collectors, and factories retain MySQL behavior by default, and external database support remains explicitly opt-in.

Generated with Codex


aparajon (approved):
🤖 Reviewed on Armand's behalf — full read of the diff plus a local run of the entire suite with `-race` against the MySQL 8.0/8.4 test containers: all green.

What I verified:
- **Registry + config gating fail closed.** `RegisterDatabaseModule` rejects nil/`mysql`/`*`/invalid names; external database types reject MySQL-only config (socket, my.cnf, heartbeat, exporter, `plans.table`) and require a registered module; `Redacted()` nils the opaque `DatabaseConfig` so secrets can't leak through config dumps.
- **Credentials parity.** The new engine-neutral `credentials` package preserves the existing MySQL precedence (IAM → Secrets Manager → password file → static), including the IAM default-port behavior in `dbconn`.
- **Plan/monitor compatibility.** DB-type validation runs both at plan load and at `Plan()` selection, so a shared plan can't silently attach to an incompatible monitor.
- **Monitor lifecycle.** The generation-scoped `stopRun` means a stale subsystem can't tear down a newer monitor generation, and `closeDB()` on stop/restart fixes a pre-existing `*sql.DB` leak on subsystem restart. The engine's `collectorWG` join plus `Exporter.Stop()` close the remaining cleanup gaps. I also walked the stop path for deadlocks (wg.Wait under runMux vs the run goroutine's LIFO defers) — it's sound.

Two non-blocking notes:
- `interpolateEnv`/`interpolateMon` switching to `ReplaceAllStringFunc` is a quiet bugfix: previously a value with multiple placeholders had every match replaced with the first var's value, and `$` in replacement values could be regex-expanded. Strictly more correct, but a config relying on the old quirk would change behavior.
- `loader.monitorTypes` entries are added on `LoadMonitor` but never pruned when a monitor goes away, so a long-lived loader's shared-plan validation set can only grow. Harmless today; worth a TODO if monitor churn ever matters.

*Reviewed by Armand's AI agent (Claude Fable 5).*

